PR #4, opened 2020-04-01
view, create, edit hospitals

petorsong:


src/components/CreateHospital.js
```diff
@@ -0,0 +1,55 @@
...
+        addHospital(name: $name){
+            id
+            name
+            createdAt
+            updatedAt
+        }
+    }
+`;
+
+class CreateHospital extends Component {
```
JasonYG, line 17: I think it might be better practice if we stuck to functional components instead of class components. App is already a functional component, and we could also use the new React hooks. Apollo client has some nice support for React hooks - they wrote their own `useQuery`, and other hooks for CRUD operations.

src/components/CreateHospital.js
```diff
@@ -0,0 +1,55 @@
...
+                <div>
+                    <input
+                        className="mb2"
+                        value={name}
+                        onChange={e => this.setState({ name: e.target.value })}
+                        type="text"
+                        placeholder="Hospital Name"
+                    />
+                </div>
+                <Mutation
```
JasonYG, line 35: I haven't looked into this way of writing mutations much yet, but I think the newer approach is to use the React hook `useMutation`. See [here](https://www.apollographql.com/docs/react/data/mutations/) for more info. I'm not sure what the pros and cons of both methods are - I'll look into that - but it seems that the React hooks might be better practice, it's definitely newer.

src/components/Hospital.js
```diff
@@ -0,0 +1,18 @@
+import React, { Component } from 'react';
+import { Link } from 'react-router-dom';
+
+class Hospital extends Component {
+    render() {
+        const { id, name, createdAt, updatedAt } = this.props.hospital;
+        return (
+            <tr>
+                <td><Link to={{ pathname: "/editHospital", state: {hospital: this.props.hospital}}}>{id}</Link></td>
+                <td>{name}</td>
```
JasonYG, line 10: Right now, we're hard coding in the properties we want displayed (id, name, createdAt, updatedAt), but a better way might be to just map all the properties of `this.props.hospital`.

JasonYG: so maybe something like:

`{this.props.hospital.map(property => <td>{property}</td>)}`

JasonYG: sorry i'm not sure how to directly suggest changes on the code

JasonYG: I'm not sure if this is the intended functionality though - let me know what you think.

package.json
```diff
@@ -13,11 +13,16 @@
...
+    "graphql": "^14.6.0",
     "react": "^16.13.1",
+    "react-apollo": "^3.1.3",
     "react-dom": "^16.13.1",
+    "react-router": "^5.1.2",
+    "react-router-dom": "^5.1.2",
     "react-scripts": "3.4.1",
-    "serve": "^11.3.0"
+    "serve": "^11.3.0",
+    "typescript": "^2.8.0"
```
JasonYG, line 25: I noticed we have typescript here - will we be using typescript to build our components?

src/components/App.js
```diff
@@ -0,0 +1,24 @@
...
+import CreateHospital from "./CreateHospital";
+import EditHospital from "./EditHospital";
+
+function App() {
+  return (
+      <div className="center w85">
+        <Header />
+        <div className="ph3 pv1 background-gray">
+          <Switch>
+            <Route exact path="/viewHospitals" component={HospitalTable} />
```
JasonYG, line 15: Once we start building out the app, maybe we should nest these all under one route - for example, `/hospitals/view`, `/hospitals/edit`, `/hospitals/create`, etc

package.json
petorsong: not sure how it sneaked in here but i'm considering it - i don't think anyone on our team has much experience which is what's holding me back

src/components/App.js
petorsong: that's a good idea for sure, but this will also depend on the structure of our app in general - i'm leaning towards full screen modals instead of specific routes as a better solution

src/components/CreateHospital.js
petorsong: yep, we should definitely move towards hooks and functional components - this is probably one of the first changes i'd like to make to what's here right now

src/components/CreateHospital.js
petorsong: i haven't used hooks too much but i'll look into it, does seem to be more widely used nowadays

src/components/Hospital.js
petorsong: yeah this is with my limited knowledge of `react-router`, but you're right, spreading the properties on the component would be better in general

PR #8, opened 2020-06-11
Add New Event Flow

JasonYG:
## Overview
Created the add new event flow for creating an event.

Note: the maps/location feature is NOT working as of now

[Notion task](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=8218077cb67143e2a5ec37b2450b1e78)

DESCRIPTION

## Changes
- Added new route and page, `/events/new`, for event creation flow
- Created mutation to the backend to add new event

## Screenshots
Screen Recording 2020-06-09 at 6 12 13 PM

## Testing
- Verified that the mutation worked

## Checklist
- [ ] Change title to be more descriptive (shouldn't be just branch name)
- [ ] Update `Notion task` link
- [ ] `Overview`
- [ ] `Changes`
  - [ ] \(Optional) Add inline comments in diff
- [ ] \(Optional) `Screenshots`
- [ ] \(Optional) `Testing` (N/A if nothing changed)
- [ ] \(Optional) `Other Notes`
- [ ] Request at least 2 reviewers
- [ ] Assign yourself to the PR
- [ ] Make sure build passes

@uwblueprint/paramedics


rickymao (commented):
Everything looks great, however when I test it the event cards are not there. I can still add a new event and the flow works. When I set up it on my side, I  just followed the builder commands and ran docker. Is there something I am missing?

JasonYG:
> Everything looks great, however when I test it the event cards are not there. I can still add a new event and the flow works. When I set up it on my side, I just followed the builder commands and ran docker. Is there something I am missing?

Make sure that your server (`paramedics-web`) is running on `localhost:4000`

rickymao:
> > Everything looks great, however when I test it the event cards are not there. I can still add a new event and the flow works. When I set up it on my side, I just followed the builder commands and ran docker. Is there something I am missing?
> 
> Make sure that your server (`paramedics-web`) is running on `localhost:4000`

Ok, I got the cards to show up. Does adding a card work? Currently when I add an event, the new one doesn't show up.

wongcmichelle:
When I scroll on the events page, I can see that the grey background doesn't fill the entirety of the bottom portion of the screen. In addition, the "Add Event" button isn't staying in position relative to the bottom of the visible screen. See attached screen recording. I don't believe this is the intent?
[Screen Recording 2020-06-14 at 12.33.16 PM.zip](https://github.com/uwblueprint/paramedics-react/files/4776848/Screen.Recording.2020-06-14.at.12.33.16.PM.zip)

Also looks weird on iPad (I used Chrome DevTools to view as iPad device).
[Screen Recording 2020-06-14 at 1.44.26 PM.zip](https://github.com/uwblueprint/paramedics-react/files/4776950/Screen.Recording.2020-06-14.at.1.44.26.PM.zip)

src/containers/EventCreationPage.tsx
```diff
@@ -0,0 +1,163 @@
...
+        <FormField
+          label="Date of Event:"
+          placeholder="YYYY:MM:DD"
+          onChange={handleDateChange}
+          value={
+            eventDate
+              ? `${dateParts.year}:${dateParts.month}:${dateParts.day}`
+              : ""
+          }
+          handleClick={handleOpenDateModal}
```
wongcmichelle: This might be a design question, but the date picker modal only opens on click. Let's say the user enters this field via tabbing from the previous field on the form, would we want the date picker modal to open then too? Consider passing a handler for the `onfocus` event.

src/components/EventCreationPage/SelectDateModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+  setEventDate: (date: any) => void;
+}) => {
+  const classes = useModalStyles();
+  const [date, setDate] = useState<any>(eventDate);
+  return (
+    <Modal open={open} onClose={handleClose}>
+      <Container classes={{ root: classes.root }}>
+        <Typography classes={{ root: classes.text }}>
+          Select Event Date:
+        </Typography>
```
wongcmichelle: This text alignment doesn't match the design (basing it off screenshot on the task card).

wongcmichelle:
Date confirmation dialog (https://www.notion.so/uwblueprintexecs/New-Event-Flow-8218077cb67143e2a5ec37b2450b1e78#b845ca2622fe4ae4bd7c0e1afe7b319e) is missing according to design. (Not sure if it's necessary but pointing it out.)

src/containers/EventCreationPage.tsx
```diff
@@ -0,0 +1,163 @@
...
+        setEventDate={setEventDate}
+      />
+      <Stepper
+        activeStep={activeStep}
+        nextButton={
+          <div className="next-container">
+            <NextButton
+              handleClick={activeStep < 1 ? handleNext : handleComplete}
+              disabled={eventName === "" || eventDate === null}
+            />
```
wongcmichelle: When `activeStep < 1` is `false`, you're passing the `handleComplete` which creates the event. The NextButton text should say "CREATE" (not "next"). I suggest you pass the text as a prop instead of hardcoding "NEXT".

src/components/EventCreationPage/NextButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+      variant="contained"
+      color="primary"
+      onClick={handleClick}
+      classes={{
+        root: classes.root,
+        disabled: classes.disabled,
+      }}
+      disabled={disabled}
+    >
+      Next
```
wongcmichelle: Don't hardcode this.

wongcmichelle:
Overall comments:
This is a huge PR, great work guys :) 
Some thoughts
1. Please remove any unused imports in any of the files.
2. Since this is pretty much the first PR in this repo, would it be possible to set up a linter for moving forward? (https://medium.com/javascript-scene/streamline-code-reviews-with-eslint-prettier-6fb817a6b51d#:~:text=ESLint%20performs%20automated%20scans%20of,quotes%20vs%20double%20quotes%2C%20etc.)
3. Can we define our styles at the top of a file? (e.g. define `const useEventCardStyles = makeStyles({...})` before you actually use it)
4. Want to talk about the file organization because we have files called `...Page` inside `containers` and folders called `...Page` outside of the containers folder.

src/containers/EventCreationPage.tsx
```diff
@@ -0,0 +1,163 @@
...
+  useEventMutation(
+    {
+      name: eventName,
+      eventDate: eventDate || new Date(),
+      createdBy: 1,
+      isActive: true,
+    },
+    complete,
+    setComplete
+  );
```
wongcmichelle: Is there a reason why this isn't called inside `handleComplete`? (instead of setting `complete` and checking whether `complete` is true in the function definition of `useEventMutation`)

src/containers/EventCreationPage.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  const handleComplete = () => {
+    setComplete(true);
+    window.location.href = "/events/";
```
wongcmichelle: can we try using react-router's `useHistory` hook instead? https://stackoverflow.com/questions/31079081/programmatically-navigate-using-react-router
It should make it work more like a single page application (won't refresh the page).

wongcmichelle:
> > > Everything looks great, however when I test it the event cards are not there. I can still add a new event and the flow works. When I set up it on my side, I just followed the builder commands and ran docker. Is there something I am missing?
> > 
> > 
> > Make sure that your server (`paramedics-web`) is running on `localhost:4000`
> 
> Ok, I got the cards to show up. Does adding a card work? Currently when I add an event, the new one doesn't show up.

I am also experiencing this issue. Adding an event through the UI only works sometimes. My initial guess is because `useEventMutation` is currently set up so that it relies on `complete` and `setComplete`. Setting state is asynchronous so you may be running into issues there.

src/containers/HomeLandingPage.tsx
```diff
@@ -46,15 +47,27 @@ const HomeLandingPage = () => {
...
+          spacing={3}
+        >
+          {events.map((event: EventType) => (
+            <Grid item>
+              <EventCard
+                date={event.eventDate}
+                eventTitle={event.name}
+                address="N/A"
+              />
+            </Grid>
```
wongcmichelle: Getting this console warning `index.js:1 Warning: Each child in a list should have a unique "key" prop.` Please add a `key` prop to `EventCard`. Could be `key={event.name}`.

src/components/common/MenuTabs.tsx
```diff
@@ -31,7 +31,7 @@ const MenuTabs = ({
             </Typography>
```
wongcmichelle, line 51: Console warning `Warning: Each child in a list should have a unique "key" prop.`. Please add a `key` prop to the `Tab` component like `key={label}`.

src/containers/EventCreationPage.tsx
JasonYG: Yeah, React forces us to have all custom hooks on the top-level of the component, with no condition. It's why we have the `complete` and `setComplete` conditional in the first place, although it's not ideal

JasonYG:
> Overall comments:
> This is a huge PR, great work guys :)
> Some thoughts
> 
> 1. Please remove any unused imports in any of the files.
> 2. Since this is pretty much the first PR in this repo, would it be possible to set up a linter for moving forward? (https://medium.com/javascript-scene/streamline-code-reviews-with-eslint-prettier-6fb817a6b51d#:~:text=ESLint%20performs%20automated%20scans%20of,quotes%20vs%20double%20quotes%2C%20etc.)
> 3. Can we define our styles at the top of a file? (e.g. define `const useEventCardStyles = makeStyles({...})` before you actually use it)
> 4. Want to talk about the file organization because we have files called `...Page` inside `containers` and folders called `...Page` outside of the containers folder.

1. Good catch, that's our bad. We'll double check that for all files.
2. Yeah, this would be ideal for sure. I think Peter said we'll address this in a separate ticket.
3. We opted to have our components defined at the top, rather than the styles. This way when we opened any component file, we'd see the React component right away instead of the styling. Thoughts?
4. Yeah, so we opted to have `containers` be where all the full-page views be, then we'd have all the components for those pages defined in `src/components/X_X_Page`.

wongcmichelle:
> > Overall comments:
> > This is a huge PR, great work guys :)
> > Some thoughts
> > 
> > 1. Please remove any unused imports in any of the files.
> > 2. Since this is pretty much the first PR in this repo, would it be possible to set up a linter for moving forward? (https://medium.com/javascript-scene/streamline-code-reviews-with-eslint-prettier-6fb817a6b51d#:~:text=ESLint%20performs%20automated%20scans%20of,quotes%20vs%20double%20quotes%2C%20etc.)
> > 3. Can we define our styles at the top of a file? (e.g. define `const useEventCardStyles = makeStyles({...})` before you actually use it)
> > 4. Want to talk about the file organization because we have files called `...Page` inside `containers` and folders called `...Page` outside of the containers folder.
> 
> 1. Good catch, that's our bad. We'll double check that for all files.
> 2. Yeah, this would be ideal for sure. I think Peter said we'll address this in a separate ticket.
> 3. We opted to have our components defined at the top, rather than the styles. This way when we opened any component file, we'd see the React component right away instead of the styling. Thoughts?
> 4. Yeah, so we opted to have `containers` be where all the full-page views be, then we'd have all the components for those pages defined in `src/components/X_X_Page`.

1. Thanks :)
2. Yup, I spoke with Peter about that. Thanks for clarifying.
3. It looks like it might be more standard to define styles at the top. Also, as someone reading the code it makes more sense to see the definition before its called (even if it's hoisted in typescript).
4. I don't think we need a containers folder and I noticed some inconsistencies. I hope the folder structure [here](https://www.notion.so/uwblueprintexecs/Front-End-Best-Practices-ba6081af21f14de0bfe3f7d12a757e12) makes sense. If not, feel free to shoot me a message and we can discuss a bit more :)

src/containers/EventCreationPage.tsx
wongcmichelle: I think this isn't working as you expect because setState is asynchronous and it looks like i'm not consistently hitting the endpoint when I create an event through the UI. (Events are only actually added sometimes)

wongcmichelle: Maybe you don't need a custom hook like `useEventMutation` and should directly use `useMutation` instead? (Sorry I haven't worked with graphql before but I'm basing off of the example [in the docs](https://www.apollographql.com/docs/react/api/react-hooks/#usemutation))

src/containers/EventCreationPage.tsx
carelynntsai: In the updated mockups, the next button text never changes to "CREATE" on the last step. I'm actually wondering if this was purposeful or a small error. I'll ask Gracie right now and update it when she confirms.
image

petorsong (changes requested):
running into the same issue with events not being created via the UI, as long as that's fixed i'm good with this PR - we'll need to do a styling pass but i'm going to take that out of this PR's scope as that discussion should be at a higher level

src/containers/EventCreationPage.tsx
```diff
@@ -0,0 +1,183 @@
...
+          dateParts.day &&
+          `${dateParts.year}-${dateParts.month.padStart(
+            2,
+            "0"
+          )}-${dateParts.day.padStart(2, "0")}`,
+        createdBy: 1,
+        isActive: true,
+      },
+    });
+    history.replace("/events");
```
carelynntsai, line 98: Hi @wongcmichelle, is there a better way to take the user back to /events and re-render the page so the new event appears?

wongcmichelle: What is the issue with the way you're doing it now (using `useHistory()`)?

wongcmichelle (dismissed):
Add event is now consistently working, there's no page reload, and i can see the added event without refreshing the page 💯

petorsong (dismissed):

petorsong (approved):

wongcmichelle (approved):

PR #11, opened 2020-06-28
Add event dashboard page

wongcmichelle:
## Overview
- Added Event Dashboard Page
- Connected menu to CCP Dashboard

[Notion task](https://www.notion.so/uwblueprintexecs/Event-Dashboard-Screen-af7c8340fc394304abc1cabaa5ae2296)

- TO DO (separately): Add and update `MenuAppBar` to fit in CCP Dashboard (highlight corresponding checkpoint in menu)

## Changes
- Added route for Event Dashboard Page (at `/events/{eventId}`)
-  Add Event Dashboard Page with CCP, Hospital, Ambulance tabs
    - Tables in each tab panel use real data
    - CCP table doesn't actually have addresses
    - Functionality of including/excluding multiple and single hospitals/ambulances
    - Included select all and table sort 
- Display event specific info (no addresses right now)
- Hamburger menu with contents

## Testing
1. Go to an event dashboard page via the route `/events/{eventId}`
2. Check that the event name and date are correct
3. (Add some CCPs, Hospitals, and Ambulances to the event in the backend to verify they show up and have the correct active/inactive status and include/exclude button)
4. On the Hospital and Ambulance tabs, the select all box should select all table rows and the number selected should be updated accordingly. Verify that the include/exclude buttons on each table row are disabled and that include/exclude buttons appear above the table.
5. Try including/excluding hospitals via selecting checkboxes. 
    - Check that the confirmation dialog includes the selected hospitals.
    - Cancel should close dialog and deselect
    - Confirm should include/exclude (status should update without page refresh)
6. Try including/excluding a single hospital using the button in a table row.
7. Try including/excluding ambulances via selecting checkboxes. Check that the confirmation dialog includes the selected hospitals.
8. Try including/excluding a single ambulance using the button in a table row.
9. Verify the sort functionality on the table columns work.

## Screenshots
CCP tab:
image

Hospital tab:
image

Include Hospitals confirmation dialog:
image

Exclude Hospitals confirmation dialog: (Notice the title is plural, but the question/prompt is singular "hospital")
image

Ambulance tab:
image

Include Ambulances confirmation dialog:
image

Exclude Ambulances confirmation dialog: (Notice the title is plural, but the question/prompt is singular "ambulance")
image

Hamburger menu:
image


src/components/EventDashboard/ResourceTabPanel.tsx
```diff
@@ -0,0 +1,312 @@
...
+
+  const { data } = useQuery(query);
+
+  const rows = data ? ( type === TabOptions.Hospital ? createHospitalData(data.hospitals) : createAmbulanceData(data.ambulances) ) : [];
+
+  const handleRequestSort = (event: React.MouseEvent, property: string) => {
+    const isAsc = orderBy === property && order === 'asc';
+    setOrder(isAsc ? 'desc' : 'asc');
+    setOrderBy(property);
+  };
```
wongcmichelle, line 211: oops missed this, should move this to the `src/utils/sort.ts` file

.eslintrc
```diff
@@ -36,6 +36,8 @@
     "@typescript-eslint/no-explicit-any": "off",
     "no-shadow": "off",
     "no-plusplus": "off",
-    "react/jsx-curly-newline": "off"
+    "react/jsx-curly-newline": "off",
+    "import/no-cycle": "off",
+    "no-nested-ternary": "off"
```
wongcmichelle, line 41: This is intentional

src/components/EventDashboard/ResourceTabPanel.tsx
```diff
@@ -0,0 +1,541 @@
...
+    mutationOptions
+  );
+  const [deleteHospitalsFromEvent] = useMutation(
+    DELETE_HOSPITALS_FROM_EVENT,
+    mutationOptions
+  );
+  const [deleteAmbulancesFromEvent] = useMutation(
+    DELETE_AMBULANCES_FROM_EVENT,
+    mutationOptions
+  );
```
wongcmichelle: I'm going to try to explain why I'm refetching the `GET_EVENT_BY_ID` query after each of these mutations.

This is the part of the cache that is relevant here: (courtesy of Apollo Client Devtools)
image

These mutations will automatically update `events` in the cache **IF** the event with the correct eventId (i.e. `id = 3` in this case) is existing in the `events` field of the cache. However, I'm using `GET_EVENT_BY_ID` to get event specific data, which is added to the cache in the field `event({"id":"3"})`.

IDEALLY, we would use [readQuery](https://www.apollographql.com/docs/react/caching/cache-interaction/#readquery) so that the `GET_EVENT_BY_ID` queries the cache and not the server (and thus using the automatically updated data in the cache (I hope)). UNFORTUNATELY, `readQuery` throws an error if the cache doesn't contain all of the data necessary to fulfill a specified query. It never attempts to fetch data from a remote server. As such, it's not really usable for our use case here. I also could not figure out a way to properly `update` the query (in the cache using `cache.writeQuery` or `cache.modify`) without refetching in this particular use case (specifically, querying by id). 

It also honestly seems kinda odd to me that the `GET_EVENT_BY_ID` result is cached separately and not inside the `events` field.

I've spent a long time investigating this issue, and if anyone has a better solution, I'd be happy to try it out, but this is what I ended up implementing. I'll look into it more if I have time.

wongcmichelle: SOLVED BY RETURNING ID in `GET_EVENT_BY_ID` although it still is kinda odd to me that the `GET_EVENT_BY_ID` result is cached separately and not inside the events field.

src/graphql/queries/events.ts
```diff
@@ -1,24 +1,63 @@
 import gql from 'graphql-tag';
 
 export interface EventType {
+  id: string;
   name: string;
   eventDate: Date;
+  // ambulances: Ambulance[];
+  // hospitals: Hospital[];
```
petorsong: was this intentional?

src/graphql/queries/events.ts
```diff
@@ -1,24 +1,63 @@
...
 }
 
 export const FETCH_ALL_EVENTS = gql`
```
petorsong, line 13: both this query and `GET_ALL_EVENTS` are already used on other pages - can you check that these changes don't affect them in any way

src/components/EventDashboard/ActionConfirmationDialog.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { TabOptions } from './EventDashboardPage';
+import { ResourceData, HospitalData, AmbulanceData } from './ResourceTabPanel';
+
+interface ActionConfirmationDialogProps {
+  open: boolean;
+  type: TabOptions.Hospital | TabOptions.Ambulance;
+  handleClose: () => void;
+  handleAdd: () => void;
+  handleDelete: () => void;
+  includeOrExclude: 'include' | 'exclude' | null;
```
petorsong, line 24: would this be more performant than string comparisons?
```suggestion
  isInclude: boolean;
```

src/components/EventDashboard/CCPTabPanel.tsx
```diff
@@ -0,0 +1,177 @@
...
+  buttonIcon: {
+    marginRight: '13px',
+  },
+});
+
+interface CCP {
+  id: string;
+  name: string;
+  eventId: { id: string };
+}
```
petorsong: can this be pulled out an reused elsewhere? this entity is generic/ubiquitous enough that it shouldn't live in this component (or maybe it should if we really don't use this anywhere else)

src/components/EventDashboard/ResourceTabPanel.tsx
```diff
@@ -0,0 +1,518 @@
...
+
+interface Hospital {
+  id: string;
+  name: string;
+}
+
+interface Ambulance {
+  id: string;
+  vehicleNumber: number;
+}
```
petorsong: ditto for these, maybe can be pulled out?

src/components/EventDashboard/ResourceTabPanel.tsx
```diff
@@ -0,0 +1,518 @@
...
+  TableCell,
+  TableContainer,
+  TableHead,
+  TableRow,
+  TableSortLabel,
+  Toolbar,
+} from '@material-ui/core';
+import { FiberManualRecord, Add, Remove } from '@material-ui/icons';
+import { Order, stableSort, getComparator } from '../../utils/sort';
+import { GET_EVENT_BY_ID } from '../../graphql/queries/events';
```
petorsong: missed a linting error smh
`error  'GET_EVENT_BY_ID' is defined but never used  @typescript-eslint/no-unused-vars`

```suggestion
```

petorsong (changes requested):
looks good overall, just some minor gripes

re: your comment on caching and `GET_EVENT_BY_ID` - nice deep dive into the topic, i don't have too much context into it and have my own theories, but if it's working as expected for now we can leave this as is - any additional improvements can be made when we start the offline/caching work in earnest

src/components/EventDashboard/ResourceTabPanel.tsx
carelynntsai: These could be exported from the `hospitals.ts` and `ambulances.ts` queries files

src/components/EventDashboard/CCPTabPanel.tsx
carelynntsai: There's another existing CCP interface in `queries/cpp.ts`, can that be used here instead

src/components/EventDashboard/EventDashboardPage.tsx
```diff
@@ -0,0 +1,138 @@
...
+          <Typography variant="body1" style={{ paddingRight: '52px' }}>
+            Ezra Street l1j3j4, Waterloo Canada
+          </Typography>
+          <CalendarTodayOutlinedIcon className={classes.icon} />
+          <Typography variant="body1">{formatDate(event.eventDate)}</Typography>
+        </Box>
+      </Container>
+      <Tabs
+        textColor="secondary"
+        className={classes.tabs}
```
carelynntsai, line 109: should indicator be thicker to be more consistent with other tabs designs?

carelynntsai (commented):
Looks great! 🍕

src/graphql/queries/events.ts
wongcmichelle: I commented them out because `Ambulance` and `Hospital` were undefined and I thought that would be fixed once `resource-management` was merged into master. But i'll just go ahead and add them.

src/components/EventDashboard/ActionConfirmationDialog.tsx
wongcmichelle: I don't think it makes much of a difference. Also would prefer having three distinct states because `false` and `null` can become kind of ambiguous.

src/graphql/queries/events.ts
wongcmichelle: This was intentional because `Ambulance` and `Hospital` were undefined, but I've resolved this now.

src/components/EventDashboard/EventDashboardPage.tsx
wongcmichelle: I think we'll leave this for now, because we will likely refactor all the tabs in our app to use `MenuTabs` in the `common` components folder.

src/components/EventDashboard/ActionConfirmationDialog.tsx
petorsong: technically the boolean option does give you three distinct states (true, false, null), but you're right - shouldn't really do much in terms of perf anyways

petorsong (approved):
i can merge it in manually if circleci is blocking

carelynntsai (approved):

PR #18, opened 2020-07-19
Add barcode scanning

JasonYG:
[Notion task](https://www.notion.so/uwblueprintexecs/9224caeb21ce4350b8a1dc378d393323?v=43f44ac04d734394b7f7edc636c523b4)

## Changes
- Added route to `events/:eventId/ccps/:ccpId/scan
- Added barcode scanning using a library called [QuaggaJS](https://serratus.github.io/quaggaJS/)
- Redirect to patient profile screen upon barcode scan/manual entry

## Testing
- Test out flow (e.g. CCP dashboard -> barcode scanning -> patient profile screen, barcode scanning -> manual entry -> patient profile)
- Scan code 128 barcode

Note: Should test with actual barcodes from the paramedics

## Screenshots
image
image


src/components/CCPDashboard/ScanPatientButton.tsx
```diff
@@ -19,12 +20,16 @@ const useStyles = makeStyles({
 
 export const ScanPatientButton = () => {
```
aaronabraham311, line 21: I am not familiar with best practices, but I would assume that a button that links to another page is quite common. Is it worth pulling out into `common` components instead?

aaronabraham311, line 21: I am not familiar with best practices, but I would assume that a button that links to another page is quite common. Is it worth pulling out into common components instead?

src/components/ScanPatientPage/BarcodeScan.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  useEffect(() => {
+    if (!loading && barcode !== "") {
+      const selectedPatient = data.patients.filter(
+        (patient) => patient.barcodeValue.toString() === barcode
+      );
+
+      if (selectedPatient.length > 0) {
+        // Found patient
+        const {
```
aaronabraham311, line 21: Correct me if I am wrong on this Jason, but are you unpacking the patient data to get the CCP ID over here depending on whether you can match a patient in our database to an existing barcode value?

src/components/ScanPatientPage/BarcodeScan.tsx
```diff
@@ -0,0 +1,79 @@
...
+            facing: "environment", // or user
+          },
+        },
+        decoder: {
+          readers: ["code_128_reader"],
+        },
+      },
+      function (err) {
+        if (err) {
+          console.log(err);
```
aaronabraham311: Are we not doing error handling?

src/containers/EnterBarcodePage.tsx
```diff
@@ -0,0 +1,82 @@
...
+        // Found patient
+        const {
+          collectionPointId: { id: patientCCPId },
+          id,
+        } = selectedPatient[0];
+        // Redirect to patient profile
+        history.replace(`/patients/edit/${patientCCPId}/${id}`);
+      } else {
+        // No existing patient
+        history.replace(`/patients/new/${ccpID}`);
```
aaronabraham311: How are we passing the barcode value to the new screen? The entered barcode value is not appearing. This works for the edit screen

aaronabraham311 (changes requested):
This is sick. I couldn't get it to work on my end but I think that's because of the barcode and lighting.

src/components/CCPDashboard/ScanPatientButton.tsx
petorsong: i tend to agree, but for now it's not super necessary - we can revisit this after all our flows are working correctly and we want to do cleanup/refactoring

src/components/EnterBarcodePage/CompleteBarcodeButton.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import Button from "@material-ui/core/Button";
+import { makeStyles } from "@material-ui/core/styles";
+
+const CompleteBarcodeButton = ({
```
petorsong, line 14: same goes for this, right now our approach is to create this same button for every submit flow which is super redundant, but for the sake of getting these changes merged, we'll save that for later

src/components/ScanPatientPage/BarcodeScan.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { useEffect, useState } from "react";
+import { useHistory } from "react-router-dom";
+import Quagga from "quagga";
+import { useQuery } from "react-apollo";
+import { FETCH_ALL_PATIENTS } from "../../graphql/queries/patients";
+
+import "../../styles/BarcodeScan.css";
```
petorsong: should add a note for why raw css styling is used here (i believe you're looking into a way around this as well)

src/containers/EnterBarcodePage.tsx
petorsong: this is feature breaking, good catch @aaronabraham311

petorsong (changes requested):
i think this PR still has some changes in the works, the only other thing i'd like you to do is to pull master and lint your new files

i also would like to move away from using the `containers/` directory for new code, please move all scanning related code into one folder in `components/` instead

src/components/CCPDashboard/ScanPatientButton.tsx
JasonYG: yup that'd probably be best - if we refactored all the buttons that redirect to another page, but yeah probably best for another time

src/components/ScanPatientPage/BarcodeScan.tsx
JasonYG: yeah so I'm getting the CCP ID so I can redirect to the proper edit screen for the existing barcode

src/components/ScanPatientPage/BarcodeScan.tsx
JasonYG: this was boilerplate that I copied from Quagga's documentation, tbh I don't think we need this

src/components/EnterBarcodePage/CompleteBarcodeButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+      }}
+      type="submit"
+      onClick={handleClick}
+    >
+      Next
+    </Button>
+  );
+};
+
+const useCompleteBarcodeButtonStyles = makeStyles({
```
petorsong: once linted, this will need to be moved up so it's defined before we're used - in general this is the pattern we'll start to enforce from now on

src/containers/PatientProfilePage.tsx
```diff
@@ -92,6 +99,11 @@ const PatientProfilePage = ({
       });
     }
   }, [data]);
+  useEffect(() => {
+    if (mode === "new" && barcodeValue) {
+      setFormFields({ ...formFields, barcodeValue });
+    }
+  }, []);
```
petorsong, line 106: if the dependency array is empty (e.g. `[]`), this will only run once (whenever this component does get rendered) - is this what you intended?

src/components/ScanPatientPage/BarcodeScan.tsx
petorsong: if we're gonna keep this console log here, you'll have to disable this line so the linter doesn't complain

petorsong (changes requested):
couple of high-level things
- you haven't actually linted your files - here's a proposed workflow: 
  - run `npm run lint` in your frontend container - this will deal with most easy-to-fix issues automatically
  - using the output from the above, open individual files in vscode and resolve specific issues
  - lemme know if you're unsure how to fix any of the things it catches
- after entering a barcode manually and hitting next, if it's a new barcode, it immediately creates the patient (i.e. saves) - i dunno if this is intended but i think it makes more sense if we only save/create after we fill out the rest of `PatientProfilePage` first
- the camera stays on even after you leave the scan page - this is very obvious if you use a browser as it shows you exactly when the camera is being used by the page
  - a potential solution is to stop the camera when the component isn't rendered (e.g. use `componentDidMount()` maybe?)

src/containers/PatientProfilePage.tsx
JasonYG: yeah it's intended

JasonYG:
> couple of high-level things
> 
> * you haven't actually linted your files - here's a proposed workflow:
>   
>   * run `npm run lint` in your frontend container - this will deal with most easy-to-fix issues automatically
>   * using the output from the above, open individual files in vscode and resolve specific issues
>   * lemme know if you're unsure how to fix any of the things it catches
> * after entering a barcode manually and hitting next, if it's a new barcode, it immediately creates the patient (i.e. saves) - i dunno if this is intended but i think it makes more sense if we only save/create after we fill out the rest of `PatientProfilePage` first
> * the camera stays on even after you leave the scan page - this is very obvious if you use a browser as it shows you exactly when the camera is being used by the page
>   
>   * a potential solution is to stop the camera when the component isn't rendered (e.g. use `componentDidMount()` maybe?)

- Yeah my bad with the linting, I ran that now
- What do you mean by saving the patient? It's not creating any patient, it's just redirecting to the patient profile screen
- Good catch with the camera issue, I fixed it in the latest commit

src/components/EnterBarcodePage/EnterBarcodePage.tsx
```diff
@@ -38,6 +38,7 @@ const EnterBarcodePage = ({
         history.replace(`/patients/new/${ccpID}/${barcode}`);
       }
     }
+    history.replace(`/patients/new/${ccpID}/${barcode}`);
```
petorsong, line 41: why was this line added?

petorsong (changes requested):
camera issue still persists - i get that you manually turn it off on the onclick for `ScanPatientTopBar` but this leaves out a lot of usecases (e.g. clicking the back button on the same bar, clicking the back button in the browser); instead of trying to find every usecase, we should just have a catch all that turns it off the moment we know it's not being used anymore (i.e. my `componentDidMount` suggestion)

the issue where it creates the patient immediately on scan seems to be fixed though - i think it worked out once you switched the barcode structure maybe

src/components/EnterBarcodePage/EnterBarcodePage.tsx
JasonYG: it's to handle an edge case where the backend wasn't up, or if things weren't loaded in yet

JasonYG:
> camera issue still persists - i get that you manually turn it off on the onclick for `ScanPatientTopBar` but this leaves out a lot of usecases (e.g. clicking the back button on the same bar, clicking the back button in the browser); instead of trying to find every usecase, we should just have a catch all that turns it off the moment we know it's not being used anymore (i.e. my `componentDidMount` suggestion)
> 
> the issue where it creates the patient immediately on scan seems to be fixed though - i think it worked out once you switched the barcode structure maybe

Good point actually, I added the `Quagga.stop()` stuff to the equivalent of `componentDidUnmount`, which should handle edge cases (see latest commit).

petorsong (approved):
camera behaviour looks good now

aaronabraham311 (approved):

PR #21, opened 2020-07-28
Connect auth landing page to event dashboard and fix some styling

JasonYG:
[Notion task](https://www.notion.so/uwblueprintexecs/9224caeb21ce4350b8a1dc378d393323?v=43f44ac04d734394b7f7edc636c523b4)

## Changes
- Fixed styling for the auth landing page
- Clicking on an event card now redirects to the event dashboard

## Testing
- Click on an event card, and verify that it's redirecting to the correct URL

Note: the event dashboard isn't in this PR yet, so clicking on an event card won't actually change your screen
Note2: there's a small styling bug in this PR. The Add New Event button can cover an event card if there's enough of them. This should be fixed in a future PR.

## Screenshots
Landing Page
image

### DIFF
--- a/src/components/HomeLandingPage/CardOptions.tsx
+++ b/src/components/HomeLandingPage/CardOptions.tsx
@@ -19,6 +19,7 @@ const CardOptions: CardOptions = () => {
   const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
 
   const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.stopPropagation();
     setAnchorEl(event.currentTarget);
   };
 
--- a/src/components/HomeLandingPage/HomeLandingPage.tsx
+++ b/src/components/HomeLandingPage/HomeLandingPage.tsx
@@ -4,12 +4,13 @@ import Typography from '@material-ui/core/Typography';
 import Grid from '@material-ui/core/Grid';
 import AccountCircleIcon from '@material-ui/icons/AccountCircle';
 
-import MenuTabs from '../components/common/MenuTabs';
-import AddEventButton from '../components/HomeLandingPage/AddEventButton';
-import EventCard from '../components/HomeLandingPage/EventCard';
-import useAllEvents from '../graphql/queries/hooks/events';
-import { EventType, GET_ALL_EVENTS } from '../graphql/queries/events';
-import '../styles/HomeLandingPage.css';
+import MenuTabs from '../common/MenuTabs';
+import AddEventButton from './AddEventButton';
+import EventCard from './EventCard';
+import useAllEvents from '../../graphql/queries/hooks/events';
+import { EventType, GET_ALL_EVENTS } from '../../graphql/queries/events';
+import '../../styles/HomeLandingPage.css';
+import { useHistory } from 'react-router-dom';
 
 const HomeLandingPage = () => {
   const [selectedTab, setTab] = useState(0);
@@ -19,6 +20,7 @@ const HomeLandingPage = () => {
   ) => {
     setTab(newValue);
   };
+  const history = useHistory();
 
   const tabLabels = ['Current Events', 'Archived Events'];
 
@@ -28,6 +30,7 @@ const HomeLandingPage = () => {
   // Fetch events from cache
   const { data } = useQuery(GET_ALL_EVENTS);
   const events: Array<EventType> = data ? data.events : [];
+  console.log(events);
 
   return (
     <div className="landing-wrapper">
@@ -60,6 +63,7 @@ const HomeLandingPage = () => {
                 date={event.eventDate}
                 eventTitle={event.name}
                 address="N/A"
+                handleClick={() => history.push(`/events/${event.id}`)}
               />
             </Grid>
           ))}
--- a/src/components/HomeLandingPage/HomeLandingPage.tsx
+++ b/src/components/HomeLandingPage/HomeLandingPage.tsx
@@ -57,9 +57,9 @@ const HomeLandingPage = () => {
       <div className="landing-body">
         <Grid container direction="row" alignItems="center" spacing={3}>
           {events.map((event: EventType) => (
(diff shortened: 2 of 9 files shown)


src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -28,6 +30,7 @@ const HomeLandingPage = () => {
   // Fetch events from cache
   const { data } = useQuery(GET_ALL_EVENTS);
   const events: Array<EventType> = data ? data.events : [];
+  console.log(events);
```
petorsong: can you check if the linter catches this? i thought i had this rule on

petorsong (changes requested):
- if the `useeffect` error has a really easy fix, i wouldn't mind it if it was thrown into this PR too
- the selected tab still has the old dark/black colour - is this also supposed to match the blue?

src/components/HomeLandingPage/CardOptions.tsx
```diff
@@ -19,6 +19,7 @@ const CardOptions: CardOptions = () => {
   const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
 
   const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.stopPropagation();
```
wongcmichelle, line 22: What does adding this do?

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -4,12 +4,13 @@ import Typography from '@material-ui/core/Typography';
...
-import useAllEvents from '../graphql/queries/hooks/events';
-import { EventType, GET_ALL_EVENTS } from '../graphql/queries/events';
-import '../styles/HomeLandingPage.css';
+import MenuTabs from '../common/MenuTabs';
+import AddEventButton from './AddEventButton';
+import EventCard from './EventCard';
+import useAllEvents from '../../graphql/queries/hooks/events';
+import { EventType, GET_ALL_EVENTS } from '../../graphql/queries/events';
+import '../../styles/HomeLandingPage.css';
+import { useHistory } from 'react-router-dom';
```
wongcmichelle: I'm getting a linter error for the order of the imports. `react-router-dom` import should occur before import of `../common/MenuTabs`

src/components/HomeLandingPage/HomeLandingPage.tsx
wongcmichelle: I'm getting the linter error for this console.log

wongcmichelle (changes requested):
It's working functionality wise :) Perhaps go through these files and lint them?

Also I'm assuming we're not fixing all styling issues here, but I'll point out two quick changes that would match the design better
- `indicatorColor` in `MenuTabs.tsx` should be "secondary"
- Missing some margin/bottom that prevents the "Add New Event" button from covering the event cards in the last row

src/components/HomeLandingPage/CardOptions.tsx
petorsong: i think this makes it so clicking the options won't also trigger the card onclick (which would take you to the event dashboard) - if so we should add a comment in code just to make it clear

src/components/HomeLandingPage/HomeLandingPage.tsx
JasonYG: my bad, I forgot to run the linter

src/components/HomeLandingPage/CardOptions.tsx
JasonYG: yeah Peter's right, adding this prevents the user from being redirected to the event dashboard when they click on options

JasonYG:
> It's working functionality wise :) Perhaps go through these files and lint them?
> 
> Also I'm assuming we're not fixing all styling issues here, but I'll point out two quick changes that would match the design better
> 
> * `indicatorColor` in `MenuTabs.tsx` should be "secondary"
> * Missing some margin/bottom that prevents the "Add New Event" button from covering the event cards in the last row

could you post a screenshot of the Add New Event issue? I'm not seeing the problem on my end

petorsong (dismissed):

wongcmichelle:
> > It's working functionality wise :) Perhaps go through these files and lint them?
> > Also I'm assuming we're not fixing all styling issues here, but I'll point out two quick changes that would match the design better
> > 
> > * `indicatorColor` in `MenuTabs.tsx` should be "secondary"
> > * Missing some margin/bottom that prevents the "Add New Event" button from covering the event cards in the last row
> 
> could you post a screenshot of the Add New Event issue? I'm not seeing the problem on my end

If there were one additional event card, it would be covered by the bottom. One additional styling issue. The `Tabs` component should have the prop `textColor="secondary"`.

image

src/components/common/MenuTabs.tsx
```diff
@@ -41,7 +40,7 @@ const MenuTabs: React.FC<{
   return (
     <Tabs
       value={currentTab}
-      indicatorColor="primary"
+      indicatorColor="secondary"
       textColor="inherit"
```
wongcmichelle:
```suggestion
      textColor="secondary"
```

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -54,12 +56,13 @@ const HomeLandingPage = () => {
       <div className="landing-body">
```
wongcmichelle, line 56: The css for this "landing-body" class needs `padding: 70px 56px 168px 56px;`

petorsong (approved):

petorsong:
>Note2: there's a small styling bug in this PR. The Add New Event button can cover an event card if there's enough of them. This should be fixed in a future PR.

we're gonna put this as a separate bug and deal with it in a separate ticket, since it may involve design input

wongcmichelle (approved):
LGTM

PR #23, opened 2020-08-06
CircleCI Setup

aaronabraham311:
[Notion task](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=7b3327f08a6643529ffe060e56efa83a)

## Changes
- Created CircleCI pipeline: build and install for both dev and prod environment and linting
- Linted code for video streaming 

## Testing
- Manual checks on CircleCI along with the green check on the bottom of this PR


.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "package.json" }}
+              - v1-dependencies-
+      - save_cache:
+          paths:
+            - node_modules
+          key: v1-dependencies-{{ checksum "package.json" }}
```
petorsong: normally you're supposed to do something between the `restore_cache` and `save_cache` steps, so this is actually doing nothing - i'm in favour of just removing this job entirely
```suggestion
```

.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+
+  build-prod:
+    docker:
+      - image: uwbpparamedics/uw-blueprint-paramedics-react:0.0.1
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "package.json" }}
+            - v1-dependencies-
```
petorsong: from what i understand, even if we cache this we don't do anything with it since we build the container within the `docker-compose` call
```suggestion
```

.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+      - node/with-cache:
+          steps:
+            - run: npm install
+      - setup_remote_docker:
+          docker_layer_caching: false
```
petorsong, line 11: is there a particular reason we don't call this earlier? i would've thought that this should be initialized as early as possible?

.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+            set -x
+            docker-compose -f docker-compose-prod.yaml up -d
+      - persist_to_workspace:
+          root: .
+          paths:
+            - .
+      - save_cache:
+          paths:
+            - node_modules
+          key: v1-dependencies-{{ checksum "package.json" }}
```
petorsong:
```suggestion
```

.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "package.json" }}
+              - v1-dependencies-
+      - save_cache:
+          paths:
+            - node_modules
+          key: v1-dependencies-{{ checksum "package.json" }}
```
petorsong:
```suggestion
```

.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+
+  build-prod:
```
petorsong: let's make this the convention
```suggestion
  prod-build:
```

.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "package.json" }}
+              - v1-dependencies-
+      - save_cache:
+          paths:
+            - node_modules
+          key: v1-dependencies-{{ checksum "package.json" }}
+
+  build-dev:
```
petorsong:
```suggestion
  dev-build:
```

.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+    docker:
+      - image: uwbpparamedics/uw-blueprint-paramedics-react:0.0.2
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "package.json" }}
+            - v1-dependencies-
```
petorsong:
```suggestion
```

.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+          command: touch paramedics-react.env
+      - run:
+          name: Run Docker compose
+          command: |
+            set -x
+            docker-compose -f docker-compose-dev.yaml up -d
+      - persist_to_workspace:
+          root: .
+          paths:
+            - .
```
petorsong: what does this do? is it necessary if there is no next step in this job?

.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+      - save_cache:
+          paths:
+            - node_modules
+          key: v1-dependencies-{{ checksum "package.json" }}
```
petorsong:
```suggestion
```

.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+
+  lint:
```
petorsong: just to make it super clear that we're linting in the dev container
```suggestion
  dev-lint:
```

.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+    docker:
+      - image: uwbpparamedics/uw-blueprint-paramedics-react:0.0.2
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "package.json" }}
+            - v1-dependencies-
```
petorsong:
```suggestion
```

.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+      - attach_workspace:
+          at: .
+      - run:
+          name: Lint
+          command: |
+            npm install
```
petorsong: is there a particular reason we call this specifically instead of putting it as a separate step?
e.g.
```
      - node/with-cache:
          steps:
            - run: npm install
```

.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+      - install-prod
+      - build-prod:
+          requires:
+            - install-prod
+      - build-dev:
+          requires:
+            - install-dev
+      - lint:
+          requires:
+            - build-dev
```
petorsong: if we can remove the `dev-lint` dependency for `dev-lint` that would be ideal - then we could have 3 parallels jobs running and it would be quicker overall
```suggestion
      - prod-build
      - dev-build
      - dev-lint:
        requires:
          - dev-build
```

.circleci/config.yml
```diff
@@ -1,7 +1,135 @@
...
+      - run:
+          name: Lint
+          command: |
+            npm install
+            npm run lint
+
 workflows:
-  node-tests:
+  version: 2
+  build-lint:
```
petorsong: this is what i called it but honestly not a big deal either way
```suggestion
  build-and-lint:
```

petorsong (changes requested):
left some comments based on the `config.yml` updates i made in `paramedics-web` (see the changes [in the PR](https://github.com/uwblueprint/paramedics-web/pull/60/files#diff-1d37e48f9ceff6d8030570cd36286a61) or [as a standalone file](https://github.com/uwblueprint/paramedics-web/blob/26b3c64d533803c058de6d6a935ecb5014750395/.circleci/config.yml)) but i'm open to discussion if you feel like any of these changes should be something different

.circleci/config.yml
```diff
@@ -1,7 +1,79 @@
...
+  node: circleci/node@1.1.6
+jobs:
+  prod-build:
+    docker:
+      - image: uwbpparamedics/uw-blueprint-paramedics-react:0.0.1
+    steps:
+      - checkout
+      - node/with-cache:
+          steps:
+            - run: npm install
```
petorsong: pretty sure this doesn't really benefit us as `docker-compose up` will implicitly install packages further down
```suggestion
```

.circleci/config.yml
```diff
@@ -1,7 +1,79 @@
...
+            chmod +x /usr/local/bin/docker-compose
+      - run:
+          name: Run Docker compose
+          command: |
+            set -x
+            docker-compose -f docker-compose-prod.yaml up -d
+      - persist_to_workspace:
+          root: .
+          paths:
+            - .
```
petorsong: since we don't have any jobs that depend on this one, i don't think this is necessary
reference: https://circleci.com/docs/2.0/configuration-reference/#persist_to_workspace
```suggestion
```

.circleci/config.yml
```diff
@@ -1,7 +1,79 @@
...
+
+  dev-build:
+    docker:
+      - image: uwbpparamedics/uw-blueprint-paramedics-react:0.0.2
+    steps:
+      - checkout
+      - node/with-cache:
+          steps:
+            - run: npm install
```
petorsong:
```suggestion
```

.circleci/config.yml
petorsong:
```suggestion
```

.circleci/config.yml
```diff
@@ -1,7 +1,79 @@
...
-  node-tests:
+  version: 2
+  build-and-lint:
     jobs:
-      - node/test
+      - prod-build
+      - dev-build
+      - dev-lint:
+          requires:
+            - dev-build
```
petorsong: ah i removed all the explicit `npm install` calls because i assumed we would want all jobs to run independently (as this should be more performant/faster) - i would at least try that out and compare runtimes; i'm also open to sticking with this or other options if they work out better
```suggestion
      - dev-lint
```

petorsong (changes requested):
feel free to ping my on slack if you want to discuss any of these comments

.circleci/config.yml
aaronabraham311: I tried this and this does not work. Since the npm packages or the docker image is not cached anywhere, we need to run `npm install` in order to run this in parallel.

petorsong (approved):

PR #24, opened 2020-08-09
Clean up landing page, and add archive, edit, and delete event functionality

JasonYG:
[Notion task](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=f6dbda0cc5e842eb982ea41c6f6a94dd)

## Changes
- Created common dialog component 
- Added archive event functionality
- Added edit event functionality
- Added unarchive event functionality
- Added delete event functionality
- Added route for editing events, `/events/edit/:eventId`
- Added options menu for `UserProfile` in the `HomeLandingPage`
- Added redirect to `/manage/members` for the `Resource Management` option 

## Testing
- Make sure archiving events is working, and sets the `isActive` field in the db
- Make sure the `Current Events` and `Archived Events` tab sort the events correctly
- Make sure unarchiving works
- Make sure deleting events also deletes events in your db
- Try editing an event, and see if the changes are reflected on the db level
- Try redirecting to the `Resource Management` page

## Screenshots
Archiving Events Dialog
image

Archived Events Moved to Archived Events Tab
image

Edit Event Page
image

Unarchive Event Dialog
image

Delete Event Dialog
image

User Profile Options
image

Clicking Resource Management Redirects to Here
image


src/containers/EventCreationPage.tsx
```diff
@@ -82,24 +102,59 @@ const EventCreationPage = () => {
...
+          name: eventName,
+          eventDate:
+            dateParts.year &&
+            dateParts.month &&
+            dateParts.day &&
+            `${dateParts.year}-${dateParts.month.padStart(
+              2,
+              '0'
+            )}-${dateParts.day.padStart(2, '0')}`,
+          createdBy: 16, // TODO: change this to proper user
```
petorsong: this needs to be changed back to 1 for now - all dev envs and staging guaranteed has userid 1 but not 16
```suggestion
          createdBy: 1, // TODO: change this to proper user
```

src/containers/EventCreationPage.tsx
```diff
@@ -82,24 +102,59 @@ const EventCreationPage = () => {
...
+          name: eventName,
+          eventDate:
+            dateParts.year &&
+            dateParts.month &&
+            dateParts.day &&
+            `${dateParts.year}-${dateParts.month.padStart(
+              2,
+              '0'
+            )}-${dateParts.day.padStart(2, '0')}`,
+          createdBy: 16,
```
petorsong:
```suggestion
          createdBy: 1,
```

src/containers/EventCreationPage.tsx
```diff
@@ -82,24 +102,59 @@ const EventCreationPage = () => {
...
+      }: {
+        name: string;
+        eventDate: string;
+        id: string;
+      } = data.event;
+
+      setEventName(name);
+      setEventDate(new Date(eventDate));
+    }
+  }, [data]);
```
petorsong: useEffect warning here needs to be addressed

src/graphql/queries/events.ts
```diff
@@ -34,13 +42,9 @@ export const GET_ALL_EVENTS = gql`
...
-      ambulances {
-        id
-        vehicleNumber
-      }
-      hospitals {
+      isActive
+      createdBy {
         id
-        name
       }
```
petorsong: is this query (and these deleted fields) being used somewhere else? please double check before removing fields from shared queries

src/containers/App.tsx
```diff
@@ -20,8 +20,17 @@ function App() {
...
-        <Route exact path="/events/new" component={EventCreationPage} />
+        <Route
+          exact
+          path="/events/new"
+          component={(props) => <EventCreationPage mode="new" {...props} />}
+        />
         <Route exact path="/events/:eventId" component={EventDashboardPage} />
+        <Route
+          exact
+          path="/events/edit/:eventId"
```
petorsong: we wanna follow this pattern instead
```suggestion
          path="/events/:eventId/edit"
```

src/components/common/ConfirmModal.tsx
```diff
@@ -52,10 +57,12 @@ const ConfirmModal = ({
   title,
   body,
   actionLabel,
+  actionLabelStatus = 'danger',
```
petorsong: does introducing this new prop affect existing usecases for this component (e.g. in resource management overview screens like `/manage/members`)

src/components/HomeLandingPage/EventCard.tsx
```diff
@@ -8,9 +8,14 @@ import CardOptions from './CardOptions';
 
 interface EventCardProps {
   date: Date;
+  eventId: string;
   eventTitle: string;
   address: string;
   handleClick: () => void;
+  handleArchiveEvent: () => void;
+  handleUnarchiveEvent: () => void;
+  handleDeleteEvent: () => void;
```
petorsong, line 20: minor refactoring nit - since these callbacks will only be used within this component, is it possible to set up the mutations here and just pass in the `event` object to this component? i see that `HomeLandingPage` does have a `refetch` call that might not work in this context, so if this isn't possible that's fine too

petorsong (changes requested):
works functionally but a couple of small points - would be good to design review first before making any changes that need design input
- the title of the events form accounts for if it's edit or create, but the final button is still `Create` (maybe should be `Save` for edit?)
- let's move `EventCreationPage` into its corresponding `/components` subdirectory to match everything else

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -28,24 +29,59 @@ const HomeLandingPage = () => {
...
+      variables: {
+        id: event.id,
+        name: event.name,
+        eventDate: event.eventDate,
+        createdBy: event.createdBy.id,
+        isActive: false,
+      },
+    });
+    // Force re-render
+    refetch();
```
wongcmichelle: Why are we doing all these refetches?

wongcmichelle:
Got this error after submitting an edited event :/ This was upon being redirected to `http://localhost:3000/events`. I believe this is because of [this](https://github.com/uwblueprint/paramedics-react/pull/24#discussion_r471020900) which Peter mentioned. So I'll double check this is fixed after you make the suggested changes.

image

src/containers/EventCreationPage.tsx
```diff
@@ -140,7 +195,9 @@ const EventCreationPage = () => {
     <div className="landing-wrapper">
       <div className="event-creation-top-section">
         <div className="landing-top-bar">
-          <Typography variant="h3">Create New Event</Typography>
+          <Typography variant="h3">
+            {mode === 'new' ? 'Create New Event' : 'Edit Event'}
+          </Typography>
```
wongcmichelle, line 211: On the (new) line 235, you need to update the logic since the button shouldn't say "Create" if `mode === 'edit'`

Also would suggest putting `'new'` and `'edit'` into an enum somewhere.

image

src/graphql/mutations/events.ts
```diff
@@ -19,6 +19,32 @@ export const ADD_EVENT = gql`
...
+    $isActive: Boolean!
+  ) {
+    updateEvent(
+      id: $id
+      name: $name
+      eventDate: $eventDate
+      createdBy: $createdBy
+      isActive: $isActive
+    ) {
+      name
```
wongcmichelle, line 39: You have to return the event `id` here for the cache to update automatically. Then you probably won't need to call `refetch()` like you're doing :)

src/styles/Theme.ts
```diff
@@ -90,5 +90,8 @@ export const Theme = createMuiTheme({
     MuiTab: {
       wrapper: { textTransform: 'none' },
     },
+    MuiDialogContent: {
+      root: { padding: '16px 24px' },
```
wongcmichelle: Please make sure this is what we want because it will affect all our dialogs.

wongcmichelle (changes requested):

src/containers/EventCreationPage.tsx
JasonYG: good call, will make this fix

src/graphql/mutations/events.ts
JasonYG: I added the event id here, but the cache isn't updating automatically without `refetch()`. I need to refresh my browser for the updates to show.

src/styles/Theme.ts
JasonYG: Good catch, I forgot to remove this when I refactored to the other confirmation modal component. Thanks!

src/containers/EventCreationPage.tsx
JasonYG: Good catch, forgot to include that. I'll also move 'new' and 'edit' to an enum

src/components/HomeLandingPage/HomeLandingPage.tsx
JasonYG: See below, but including the id of the event in the `editEvent` mutation didn't update the cache automatically. I included the refetches so that the event cards update on the client side without the user refreshing their browser

src/components/HomeLandingPage/EventCard.tsx
JasonYG: hmm yeah I could move these mutations to the EventCard component, but then I'd need to pass in the `event` object and the `refetch()` function. I'm thinking I should keep the mutation operations out of the EventCard component as much as possible, to make it as 'pure' of a component as possible. Any thoughts?

src/components/common/ConfirmModal.tsx
JasonYG: No it shouldn't, I made it so that existing uses of the component default to what it was before (the `danger` status on the button)

src/containers/App.tsx
JasonYG: I'll make this change

src/graphql/queries/events.ts
JasonYG: Yeah good catch, I'm not sure why I removed this lol I'll add it back

src/containers/EventCreationPage.tsx
JasonYG: good catch I didn't see this, I'll add in `loading` and `mode` to the dependency array (although this shouldn't make a difference much)

wongcmichelle:
Comment about the archive/unarchive/edit/delete popover on an event card: 

The popover should automatically close when you confirm archive/unarchive/delete from the confirmation dialog without requiring the user to click anywhere outside of the popover to "force" it to close.
image

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -40,24 +42,59 @@ const HomeLandingPage = () => {
...
```
wongcmichelle: You don't need this refetch, you need to return `isActive` in `EDIT_EVENT`.

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -40,24 +42,59 @@ const HomeLandingPage = () => {
...
+      variables: {
+        id: event.id,
+        name: event.name,
+        eventDate: event.eventDate,
+        createdBy: event.createdBy.id,
+        isActive: true,
+      },
+    });
+    // Force re-render
+    refetch();
```
wongcmichelle: Same here, you don't need this refetch, you need to return `isActive` in `EDIT_EVENT`.

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -40,24 +42,59 @@ const HomeLandingPage = () => {
...
+  };
+
+  const handleDeleteEvent = async (event: Event) => {
+    await deleteEvent({
+      variables: {
+        id: event.id,
+      },
+    });
+    // Force re-render
+    refetch();
```
wongcmichelle: I would suggest updating the cache instead of refetching data. There's examples of this in resource management (e.g. `AmbulanceOverviewPage.tsx`) https://www.apollographql.com/docs/react/data/mutations/#making-all-other-cache-updates

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -40,24 +42,59 @@ const HomeLandingPage = () => {
...
+  };
+
+  // Filters for inactive or active events
+  const filteredEvents = events.filter(
```
wongcmichelle: Should we memoize this value with `useMemo`?

src/components/common/ConfirmModal.tsx
```diff
@@ -38,6 +38,11 @@ const dialogStyles = makeStyles({
     height: 48,
     width: 107,
   },
+  dialogConfirm: {
+    color: Colours.Secondary,
+    height: 48,
+    width: 107,
```
wongcmichelle: Is this the same as `px`?

src/graphql/mutations/events.ts
```diff
@@ -20,6 +20,33 @@ export const ADD_EVENT = gql`
...
+      name: $name
+      eventDate: $eventDate
+      createdBy: $createdBy
+      isActive: $isActive
+    ) {
+      id
+      name
+      eventDate
+    }
+  }
```
wongcmichelle: You need to return `isActive` in here so that the cache automatically update when you archive/unarchive events (i.e. no refetch and no page refresh required). I would suggest returning all values to avoid similar issues in the future when someone else tries to write similar code.

wongcmichelle (changes requested):

petorsong:
> 
> 
> Comment about the archive/unarchive/edit/delete popover on an event card:
> 
> The popover should automatically close when you confirm archive/unarchive/delete from the confirmation dialog without requiring the user to click anywhere outside of the popover to "force" it to close.
> image

this is actually fixed in ricky's PR https://github.com/uwblueprint/paramedics-react/pull/26 - would be useful to review it and apply the changes here as well (one option is to merge that PR in first so you can just use the new `OptionPopper` component directly)

src/components/HomeLandingPage/HomeLandingPage.tsx
JasonYG: good suggestion!

src/components/common/ConfirmModal.tsx
JasonYG: not too sure, I just copied from above lol

JasonYG:
> > Comment about the archive/unarchive/edit/delete popover on an event card:
> > The popover should automatically close when you confirm archive/unarchive/delete from the confirmation dialog without requiring the user to click anywhere outside of the popover to "force" it to close.
> > image
> 
> this is actually fixed in ricky's PR #26 - would be useful to review it and apply the changes here as well (one option is to merge that PR in first so you can just use the new `OptionPopper` component directly)

Yup it was an easy fix - I have it so that the options menu closes whenever you open one of the confirmation dialogs.

petorsong (changes requested):
when i try to archive any event, i'm getting an error `Unhandled Rejection (Error): GraphQL error: Invalid event ID: undefined` with the following (abridged) stack trace
image

petorsong:
> 
> 
> when i try to archive any event, i'm getting an error `Unhandled Rejection (Error): GraphQL error: Invalid event ID: undefined` with the following (abridged) stack trace
> image

this has been resolved - i introduced a typo in our backend so pull master in `paramedics-web` and it should be fixed now

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -40,24 +43,71 @@ const HomeLandingPage = () => {
...
+  const [deleteEvent] = useMutation(DELETE_EVENT, {
+    update(cache, { data: { deleteEvent } }) {
+      if (!deleteEvent) {
+        return;
+      }
+      let { events } = cache.readQuery<any>({
+        query: FETCH_ALL_EVENTS,
+      });
+
+      events = events.filter((event) => event.id !== selectedEvent!.id);
```
petorsong: this line has a linter warning with the following suggested fix - just double check if this does the same as what you originally intended
```suggestion
      events = events.filter((event) => event.id !== selectedEvent?.id);
```

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -40,24 +43,71 @@ const HomeLandingPage = () => {
...
+      cache.writeQuery({
+        query: FETCH_ALL_EVENTS,
+        data: { events },
+      });
+    },
+  });
   const events: Array<Event> = data ? data.events : [];
 
+  const handleArchiveEvent = async (event: Event) => {
+    await editEvent({
```
petorsong: `await` statements for these handle- callbacks seem a little inelegant - what are they used for right now and can they be potentially refactored (e.g. with promise chaining)?

wongcmichelle: Perhaps the use of `await` can be replaced by `onCompleted` and `onError` options on the mutation.

https://www.apollographql.com/docs/react/data/mutations/#options

JasonYG: good call - those `await`'s were there back when I refetched events, but I can take those out now

JasonYG: except for `deleteEvent`, which I refactored to using `useEffect`

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -40,24 +43,76 @@ const HomeLandingPage = () => {
...
   const events: Array<Event> = data ? data.events : [];
 
+  useEffect(() => {
+    if (selectedEvent) {
+      deleteEvent({
+        variables: {
+          id: selectedEvent?.id,
+        },
+      });
+    }
```
wongcmichelle: Curious, what is your reasoning for doing this in `selectedEvent`? This doesn't really seem like the place for a `delete`. Also it looks like `selectedEvent` is only being used for delete. Can you name it more clearly if that is the case? Something like `eventToBeDeleted`?

wongcmichelle (changes requested):

src/components/HomeLandingPage/HomeLandingPage.tsx
JasonYG: Yeah good point about the naming - I'll rename it to `eventToBeDeleted` for clarity. The reason why I need this is because I needed `selectedEvent` to be set before I called the `deleteEvent` mutation - so I added it to this `useEffect` call. The async/await I had before was for this purpose, and I refactored it what it is now.

JasonYG: `selectedEvent` is used in the `deleteEvent` callback to properly update the cache

src/graphql/queries/events.ts
petorsong: @JasonYG these haven't been added back?

src/components/common/ConfirmModal.tsx
petorsong: @JasonYG please convert all values without units to `px` to keep things consistent

petorsong (changes requested):
final minor changes

src/graphql/queries/events.ts
JasonYG: I did add them back, not sure why it's not displaying properly here

src/components/common/ConfirmModal.tsx
JasonYG: i just verified that all the numbers without units default to `px`.

src/components/HomeLandingPage/UserProfile.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const classes = useStyles();
+  return (
+    <div
+      className={classes.iconWrapper}
+      onKeyDown={handleClick}
+      onClick={handleClick}
+      tabIndex={0}
+      role="button"
+    >
+      <Typography variant="h6" align="right" style={{ marginRight: '0.5em' }}>
```
petorsong: this should be changed to `px` as well

src/components/common/ConfirmModal.tsx
petorsong: let's add the units in for the sake of consistency please - for all values that don't have units especially in this file (not just these 2)
```suggestion
    height: '48px',
    width: '107px',
```

src/graphql/queries/events.ts
petorsong: can you double check your local file and make sure they're added back? as far as i can tell it's removed on this remote branch right now

petorsong (changes requested):

src/graphql/queries/events.ts
JasonYG: yup - for some reason, i think you're viewing an older commit. Try looking here: https://github.com/uwblueprint/paramedics-react/pull/24/files

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -40,24 +43,76 @@ const HomeLandingPage = () => {
...
 
+  useEffect(() => {
+    if (eventToBeDeleted) {
+      deleteEvent({
+        variables: {
+          id: eventToBeDeleted?.id,
+        },
+      });
+    }
+  }, [eventToBeDeleted]);
```
wongcmichelle: @JasonYG I still don't think that it's a good idea to do this in `useEffect`... I think `deleteEvent` should actually be called in `handleDeleteEvent`. You can set `eventToBeDeleted` when the user opens the "popper" with the options for the particular event card.

wongcmichelle (changes requested):

src/components/HomeLandingPage/UserProfile.tsx
JasonYG: good catch!

src/components/HomeLandingPage/HomeLandingPage.tsx
JasonYG: Yeah I'd agree, having this in `useEffect` doesn't seem like the best way to do this, but setting `eventToBeDeleted` when the "popper" is opened requires passing the state setter down two components. I feel like the original solution I had with async/await is prob best here tbh

src/graphql/queries/events.ts
petorsong: i'm certain i'm looking at the most up-to-date changes for this PR - even pulled it in locally and i see the same thing here
https://github.com/uwblueprint/paramedics-react/blob/42a6c89ba67d64e154404d04e6d8e79971abd53c/src/graphql/queries/events.ts#L39

regardless we can take a look at this together after standup

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -1,13 +1,15 @@
-import React, { useState } from 'react';
-import { useQuery } from 'react-apollo';
+import React, { useState, useEffect } from 'react';
```
wongcmichelle: Linter error: `useEffect` is unused

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -40,24 +43,71 @@ const HomeLandingPage = () => {
   useAllEvents();
 
   // Fetch events from cache
-  const { data } = useQuery(GET_ALL_EVENTS);
+  const { data } = useQuery(FETCH_ALL_EVENTS);
```
wongcmichelle: Because you're using `FETCH_ALL_EVENTS` here instead of `GET_ALL_EVENTS`, we no longer see the newly added event without refreshing the page. (Since in `EventCreationPage.tsx`, we only write to the cache for the `GET_ALL_EVENTS` query.)

wongcmichelle (changes requested):
Thanks for making all these changes! 2 main things:

1. We should be consistent in using either `GET_ALL_EVENTS` or `FETCH_ALL_EVENTS` (not sure what the argument here is for `FETCH_ALL_EVENTS` but I'm open to it.)
2. You mention "sort the events correctly" in the PR description. What is the correct sorting? I'm noticing when I edit an event, the card for that event remains in the same order on the events landing page. However, when I refresh the page, it moves to the very bottom of the "list" of event cards.

src/components/HomeLandingPage/HomeLandingPage.tsx
wongcmichelle: If using `GET_ALL_EVENTS` is causing any issues, I'm just going to say let's take out the `@client` and add all the fields in `FETCH_ALL_EVENTS` to the existing `GET_ALL_EVENTS` query. @petorsong thoughts?

src/graphql/queries/events.ts
```diff
@@ -34,11 +42,8 @@ export const GET_ALL_EVENTS = gql`
       id
       name
       eventDate
-      ambulances {
-        id
-        vehicleNumber
-      }
-      hospitals {
```
petorsong: i actually meant these ones should be added back too

src/components/HomeLandingPage/HomeLandingPage.tsx
petorsong: since we don't understand what `@client` does right now, i'm fine with just having 1 query - it's currently too confusing to maintain both

JasonYG:
> Thanks for making all these changes! 2 main things:
> 
> 1. We should be consistent in using either `GET_ALL_EVENTS` or `FETCH_ALL_EVENTS` (not sure what the argument here is for `FETCH_ALL_EVENTS` but I'm open to it.)
> 2. You mention "sort the events correctly" in the PR description. What is the correct sorting? I'm noticing when I edit an event, the card for that event remains in the same order on the events landing page. However, when I refresh the page, it moves to the very bottom of the "list" of event cards.

1. Good point - I'd agree with Peter, we should probably stick with just one. `GET_ALL_EVENTS` was causing errors before, likely because of the `@client` directive, so I'm thinking we should just take it out.

2. By "sort the events correctly", I just meant that the `Archive` and `Current Events` tabs displayed the right event, so only archived events would show under the `Archive` tab, and vice versa. My bad with the wording lol

src/components/HomeLandingPage/HomeLandingPage.tsx
JasonYG: yeah I'm just going to get rid of `FETCH_ALL_EVENTS` and change it to the `GET_ALL_EVENTS` query.

src/graphql/queries/events.ts
JasonYG: this PR is being kind of weird, I see those fields still exist locally. I'll push again

JasonYG: here's the file: https://github.com/uwblueprint/paramedics-react/blob/bb54f43240da91fb8820d2fdeabf65103306780f/src/graphql/queries/events.ts

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -1,13 +1,15 @@
 import React, { useState } from 'react';
-import { useQuery } from 'react-apollo';
+
```
petorsong:
```suggestion
```

src/containers/App.tsx
```diff
@@ -37,6 +42,11 @@ function App() {
           />
           <Route exact path="/events/new" component={EventCreationPage} />
```
petorsong: looks like you introduced a duplicate above
```suggestion
```

petorsong (changes requested):

src/containers/App.tsx
```diff
@@ -27,7 +27,12 @@ function App() {
           <Route exact path="/events" component={HomeLandingPage} />
           <Route
             exact
-            path="/events/:eventId/ccps/new"
+            path="/events/new"
+            component={(props) => <EventCreationPage mode="new" {...props} />}
+          />
+          <Route
+            exact
+            path="/events/:eventID/ccps/new"
```
wongcmichelle: There shouldn't be an `eventID` with this casing.
```suggestion
            path="/events/:eventId/ccps/new"
```

wongcmichelle (commented):
Small comment, otherwise LGTM :)

wongcmichelle (dismissed):
LGTM :) 🚀

petorsong (dismissed):

petorsong (approved):
cleaned up `ConfirmModal` - carelynn also made similar changes that were merged into master already; any other issues with this component is gonna be thrown into [this ticket](https://www.notion.so/uwblueprintexecs/dialog-modal-cleanup-4e0ec85b6e2f41ef8829c53a8a0fa134)

@JasonYG feel free to merge this in once @wongcmichelle gives the go ahead :shipit:

src/components/common/ConfirmModal.tsx
```diff
@@ -45,8 +40,8 @@ const dialogStyles = makeStyles({
   },
   dialogButton: {
     color: Colours.Secondary,
-    height: 48,
-    minWidth: 107,
+    height: '48px',
+    minWidth: '107px',
```
wongcmichelle, line 44: Small comment: Is it weird that this here is using the `minWidth` property while the `dialogDelete` and `dialogConfirm` styles use `width`?

wongcmichelle (approved):
Small comment, otherwise LGTM

wongcmichelle:
Not sure if "Delete" is supposed to be red.
image

src/components/common/ConfirmModal.tsx
petorsong: we can leave it in for now - when we get around to cleaning up modal styling, we can address it then

([added to this ticket](https://www.notion.so/uwblueprintexecs/dialog-modal-cleanup-4e0ec85b6e2f41ef8829c53a8a0fa134))

petorsong:
> 
> 
> Not sure if "Delete" is supposed to be red.
> image

same deal, we can address this with further cleanup

PR #28, opened 2020-08-17
Highlight new added event

wongcmichelle:
[Notion task](https://www.notion.so/uwblueprintexecs/on-entity-create-event-snackbar-4b64273fa7aa423d8504aada96547f28)

## Changes
- Highlight event card of newly added event

## Testing
- Add a new event from `/events`
- After clicking "Create" at the end of the form, it should redirect you to `/events` and the event card for the event that was added should have a blue background instead of white 
- Also try refreshing the page and navigating away from `/events` and confirm that the event card is no longer blue

## Screenshots
image

### DIFF
--- a/src/components/HomeLandingPage/EventCard.tsx
+++ b/src/components/HomeLandingPage/EventCard.tsx
@@ -1,30 +1,42 @@
 import React from 'react';
+import clsx from 'clsx';
 import Card from '@material-ui/core/Card';
 import CardContent from '@material-ui/core/CardContent';
 import Typography from '@material-ui/core/Typography';
 import Box from '@material-ui/core/Box';
 import { makeStyles } from '@material-ui/core/styles';
 import CardOptions from './CardOptions';
+import { Colours } from '../../styles/Constants';
 
 interface EventCardProps {
   date: Date;
   eventTitle: string;
+  isNew: boolean;
   address: string;
   handleClick: () => void;
 }
 
-type EventCard = ({ date, eventTitle, address }: EventCardProps) => JSX.Element;
+type EventCard = ({
+  date,
+  eventTitle,
+  isNew,
+  address,
+}: EventCardProps) => JSX.Element;
 
 const useEventCardStyles = makeStyles({
   root: {
     display: 'inline-block',
     boxShadow: 'none',
-    width: '20rem',
-    height: '10rem',
+    width: '339px',
+    height: '161px',
     cursor: 'pointer',
+    border: `1px solid ${Colours.BorderLightGray}`,
+  },
+  highlighted: {
+    backgroundColor: Colours.Blue,
   },
   cardContent: {
-    padding: '2em 2em',
+    padding: '24px 32px',
   },
   eventTitle: {
     whiteSpace: 'nowrap',
@@ -36,21 +48,26 @@ const useEventCardStyles = makeStyles({
 const EventCard: EventCard = ({
   date,
   eventTitle,
+  isNew,
   address,
   handleClick,
 }: EventCardProps) => {
   const classes = useEventCardStyles();
   return (
-    <Card className={classes.root} onClick={handleClick}>
+    <Card
+      className={clsx({
+        [classes.root]: true,
+        [classes.highlighted]: isNew,
+      })}
+      onClick={handleClick}
+    >
       <CardContent className={classes.cardContent}>
         <Box display="flex">
           <Typography color="textSecondary">{date}</Typography>
-
           <Box ml="auto">
             <CardOptions />
           </Box>
         </Box>
-
         <Typography
           color="textPrimary"
           variant="h4"
--- a/src/components/HomeLandingPage/HomeLandingPage.tsx
+++ b/src/components/HomeLandingPage/HomeLandingPage.tsx
@@ -15,7 +15,7 @@ type LocationState = { addedEventId: string | null };
 const HomeLandingPage = () => {
   const history = useHistory();
   const location = useLocation<LocationState>();
-  const { addedEventId } = location.state || { addEventId: null };
+  const { addedEventId } = location.state || { addedEventId: null };
   const [selectedTab, setTab] = useState(0);
 
   // Clear the addedEventId in location state now that it's been used
--- a/src/components/HomeLandingPage/HomeLandingPage.tsx
+++ b/src/components/HomeLandingPage/HomeLandingPage.tsx
@@ -3,8 +3,7 @@ import { useQuery } from 'react-apollo';
 import Typography from '@material-ui/core/Typography';
 import Grid from '@material-ui/core/Grid';
 import AccountCircleIcon from '@material-ui/icons/AccountCircle';
-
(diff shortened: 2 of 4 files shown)


src/components/HomeLandingPage/EventCard.tsx
```diff
@@ -1,30 +1,42 @@
...
 const useEventCardStyles = makeStyles({
   root: {
     display: 'inline-block',
     boxShadow: 'none',
-    width: '20rem',
-    height: '10rem',
+    width: '339px',
+    height: '161px',
     cursor: 'pointer',
+    border: `1px solid ${Colours.BorderLightGray}`,
```
wongcmichelle, line 33: Design asked to add this border

src/components/HomeLandingPage/EventCard.tsx
```diff
@@ -36,21 +48,26 @@ const useEventCardStyles = makeStyles({
...
   handleClick,
 }: EventCardProps) => {
   const classes = useEventCardStyles();
   return (
-    <Card className={classes.root} onClick={handleClick}>
+    <Card
+      className={clsx({
+        [classes.root]: true,
+        [classes.highlighted]: isNew,
+      })}
```
wongcmichelle, line 61: This combines the classes

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -1,26 +1,38 @@
...
+type LocationState = { addedEventId: string | null };
+
 const HomeLandingPage = () => {
+  const history = useHistory();
+  const location = useLocation<LocationState>();
+  const { addedEventId } = location.state || { addEventId: null };
   const [selectedTab, setTab] = useState(0);
+
+  // Clear the addedEventId in location state now that it's been used
+  window.history.pushState(
```
wongcmichelle, line 22: We need this so that the new event card won't be highlighted anymore after page refresh / navigating away from the page

src/containers/EventCreationPage.tsx
```diff
@@ -28,6 +28,12 @@ const EventCreationPage = () => {
         data: { events: events.concat([addEvent]) },
       });
     },
+    onCompleted({ addEvent }) {
+      history.replace('/events', { addedEventId: addEvent.id });
+    },
+    onError() {
+      history.replace('/events');
```
wongcmichelle, line 35: We might want more to be done in `onError`, but for now I included the "default" behaviour which just navigates the user to the `/events` page without setting an `addedEventId` value in the location state

src/containers/EventCreationPage.tsx
```diff
@@ -28,6 +28,12 @@ const EventCreationPage = () => {
         data: { events: events.concat([addEvent]) },
       });
     },
+    onCompleted({ addEvent }) {
+      history.replace('/events', { addedEventId: addEvent.id });
```
wongcmichelle, line 32: This adds `addedEventId` such that it can be accessed in `location.state` from anywhere in our application. Can be accessed using the `useLocation` hook.

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -1,26 +1,38 @@
...
 import useAllEvents from '../../graphql/queries/hooks/events';
 import { Event, GET_ALL_EVENTS } from '../../graphql/queries/events';
 import '../../styles/HomeLandingPage.css';
 
+type LocationState = { addedEventId: string | null };
+
 const HomeLandingPage = () => {
+  const history = useHistory();
+  const location = useLocation<LocationState>();
+  const { addedEventId } = location.state || { addEventId: null };
```
petorsong: is this a minor typo? although i can see how this does achieve the same result
```suggestion
  const { addedEventId } = location.state || { addedEventId: null };
```

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -1,26 +1,38 @@
...
   const [selectedTab, setTab] = useState(0);
+
+  // Clear the addedEventId in location state now that it's been used
+  window.history.pushState(
+    {
+      ...location.state,
+      addEventId: null,
```
petorsong: same thing here (maybe?) - kinda confusing since the key for the history state is `addedEventId` as defined in `EventCreationPage` so we _shouldn't_ have a `addEventId` in this file/context
```suggestion
      addedEventId: null,
```

petorsong (changes requested):
i believe i caught a minor typo, but otherwise this looks great! do you feel like this would be easy to extend to CRUD operations in our other tables (e.g. patients in CCP dashboard, CCPs in event dashboard, etc.)?

src/components/HomeLandingPage/HomeLandingPage.tsx
aaronabraham311: +1 to this

src/components/HomeLandingPage/HomeLandingPage.tsx
```diff
@@ -1,26 +1,38 @@
...
 import EventCard from './EventCard';
 import useAllEvents from '../../graphql/queries/hooks/events';
 import { Event, GET_ALL_EVENTS } from '../../graphql/queries/events';
 import '../../styles/HomeLandingPage.css';
 
+type LocationState = { addedEventId: string | null };
+
 const HomeLandingPage = () => {
+  const history = useHistory();
+  const location = useLocation<LocationState>();
```
aaronabraham311, line 17: Can you explain why we use `useLocation<LocationState>();` just for my understanding? Why did you not use `useState()`?

aaronabraham311 (changes requested):
Just a small clarification about the usage of `useLocation()` hook and some typos that Peter caught.

src/components/HomeLandingPage/HomeLandingPage.tsx
wongcmichelle: Ooops thanks :)

wongcmichelle:
> i believe i caught a minor typo, but otherwise this looks great! do you feel like this would be easy to extend to CRUD operations in our other tables (e.g. patients in CCP dashboard, CCPs in event dashboard, etc.)?

Thanks for catching the typo! I think this idea of passing the id in location state should be easily extendable to CRUD elsewhere :)

src/components/HomeLandingPage/HomeLandingPage.tsx
wongcmichelle: Sure, so `useState()` is used for state within a component. But I'm using `useLocation()` here to effectively pass `state` (which is part of `location`) between components upon url navigation changes. Hope this helps: https://reactrouter.com/web/api/Hooks/uselocation

petorsong (approved):

src/components/HomeLandingPage/HomeLandingPage.tsx
wongcmichelle: Let me know if you want me to go anymore in depth :)

aaronabraham311 (approved):

PR #31, opened 2020-08-23
Fix bug where CCP dashboard shows no patients

JasonYG:
[Notion task](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=d7b2b30ca27040b58fe410e7919b6655)

## Changes
- Added ID as returned field from patient queries

## Testing
- Navigate from event dashboard to CCP dashboard
- Confirm that patients correctly appear on the CCP dashboard

## Screenshots
Prior to fixing, we'd get an error from this flow:

1. From the event dashboard, navigate to a CCP dashboard
image

2. No patients show, even when they exist
image

Now, after fixing, the patients show correctly:
image

### DIFF
--- a/src/graphql/queries/patients.ts
+++ b/src/graphql/queries/patients.ts
@@ -46,6 +46,7 @@ export const GET_PATIENT_BY_ID = (id: string) => {
           id
           name
           eventId {
+            id
             name
             eventDate
           }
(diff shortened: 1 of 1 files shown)


carelynntsai (approved):
Looks good! 🍕

petorsong (approved):
will need to double check how the new ccp flow affects this but that can be done later

PR #29, opened 2020-08-17
Fixing patient routes

aaronabraham311:
[Notion task](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=8bee95fbde12454bae6646094e27ea97)

## Changes
- Made patient routes consistent with rest of the application (routes specified in the Notion ticket)
- Updated front-end mutations and queries to align with backend schema
- Updated routes for barcode components

## Testing
- Made sure routes work on browser
- Checked that data added on front end is reflected on backend

## Screenshots
Please note the routes at the top of the screenshot

Screen Shot 2020-08-24 at 1 10 05 PM

Screen Shot 2020-08-24 at 1 09 49 PM

Screen Shot 2020-08-24 at 1 31 51 PM

PR #33, opened 2020-08-24
Add ability to delete CCPs and fix adding new CCPs on event dashboard

wongcmichelle:
[Notion task 1](https://www.notion.so/uwblueprintexecs/9224caeb21ce4350b8a1dc378d393323?v=43f44ac04d734394b7f7edc636c523b4)
[Notion task 2](https://www.notion.so/uwblueprintexecs/new-ccps-missing-after-create-3885c3f1b3e848079633f6a2f862c85f)

## Changes
- Add functionality to delete CCPs
    - Confirmation modal
    - Snackbar
- Fix new CCPs not appearing in event dashboard after create
- For the sake of consistency, replace all occurrences of `eventID` with `eventId` and `ccpID` with `ccpId` in the codebase
- Small updates to `OptionPopper`
    - Remove unused classes
    - Make `Option` `styles` prop optional and combine with styles defined within the component

## Testing
Go to the event dashboard for any valid event id (`'/events/{eventId}'`)
**Deleting a CCP**
1. Click the icon button and select the "Delete" option.
    - Clicking "CANCEL" should close the "popover"
2. Delete a CCP using the icon button and click "DELETE".
    - The "popover" should close and the CCP should be removed from the list (without a refresh).
    - Refresh the page and verify that the deleted CCP does not reappear.

Additional testing to check that the CCP was removed from the cache field associated with the `GET_ALL_CCPS` query as well (requires [Apollo Client Devtools](https://github.com/apollographql/apollo-client-devtools))
1. Click "Add CCP" button then click cancel. (The CCP form page makes a `GET_ALL_CCPS` query add populates the cache with the results.)
2.  Delete a CCP using the icon button and click "DELETE" and check that the deleted CCP is removed from the cache field associated with BOTH `GET_ALL_CCPS` and `GET_CCPS_BY_EVENT_ID`.
Screen Shot 2020-08-27 at 5 42 19 PM
Screen Shot 2020-08-27 at 5 43 59 PM

**Creating a CCP**
Go to the event dashboard for any valid event id (`'/events/{eventId}'`)
1. Click "Add CCP" button. Fill in the name and click "Complete".
2. Verify that the new CCP appears in the CCP table.

## Screenshots
**Delete CCP**
image
image
image

### DIFF
--- a/src/components/CCPForm/CCPFormPage.tsx
+++ b/src/components/CCPForm/CCPFormPage.tsx
@@ -1,0 +1,181 @@
+import { useMutation } from '@apollo/react-hooks';
+import { useQuery } from 'react-apollo';
+import React, { useState, useEffect } from 'react';
+import { useHistory } from 'react-router-dom';
+import { ValidatorForm } from 'react-material-ui-form-validator';
+import Typography from '@material-ui/core/Typography';
+import { makeStyles } from '@material-ui/core';
+import Button from '@material-ui/core/Button';
+import { Colours } from '../../styles/Constants';
+import CompleteButton from './CompleteButton';
+import Map from '../EventCreationPage/Map';
+import FormField from '../common/FormField';
+import { GET_ALL_CCPS, GET_CCP_BY_ID, CCP } from '../../graphql/queries/ccps';
+import { ADD_CCP, EDIT_CCP } from '../../graphql/mutations/ccps';
+
+const useStyles = makeStyles({
+  ccpWrapper: {
+    backgroundColor: 'white',
+  },
+  ccpFormTopSection: {
+    margin: '48px 30px 0px 30px',
+    backgroundColor: 'white',
+    borderBottom: '1px solid #c4c4c4',
+  },
+  ccpHeader: {
+    display: 'flex',
+    padding: '16px 0px',
+  },
+  ccpForm: {
+    padding: '56px 56px 101px 56px',
+  },
+  ccpCompleteDiv: {
+    marginBottom: 71,
+    marginTop: 31,
+    display: 'flex',
+    justifyContent: 'flex-end',
+  },
+  ccpCancelBtn: {
+    minWidth: '107px',
+    minHeight: '48px',
+    fontSize: '18px',
+    fontWeight: 500,
+    color: Colours.Secondary,
+  },
+  ccpBtnPosition: {
+    display: 'flex',
+    marginLeft: 'auto',
+    alignSelf: 'center',
+  },
+});
+
+const CCPFormPage = ({
+  match: {
+    params: { eventID, ccpID },
+  },
+  mode,
+}: {
+  match: {
+    params: {
+      eventID: string;
+      ccpID?: string;
+    };
+  };
+  mode: string;
+}) => {
+  const classes = useStyles();
+  const history = useHistory();
+
+  const { data } = useQuery(mode === 'new' ? GET_ALL_CCPS : GET_CCP_BY_ID, {
+    variables: { id: ccpID },
+  });
+
+  const collectionPoints: Array<CCP> = data ? data.collectionPoints : [];
+  const collectionPoint: CCP = data ? data.collectionPoint : null;
+
+  const [ccpName, setCCPName] = useState<string>('');
+
+  useEffect(() => {
+    setCCPName(collectionPoint ? collectionPoint.name : '');
+  }, [collectionPoint]);
+
+  // TODO: Add actual location
+  const [eventLocation, setEventLocation] = useState<string>('');
+
+  const [addCCP] = useMutation(ADD_CCP, {
+    update(cache, { data: { addCollectionPoint } }) {
+      cache.writeQuery({
+        query: GET_ALL_CCPS,
+        data: {
+          collectionPoints: collectionPoints.concat([addCollectionPoint]),
+        },
+      });
+    },
+  });
+
+  const [editCCP] = useMutation(EDIT_CCP);
+
+  const handleNameChange = (e: any) => {
+    setCCPName(e.target.value);
+  };
+  const handleLocationChange = (e: any) => {
+    setEventLocation(e.target.value);
+  };
+
+  const handleCancel = () => {
+    history.replace(`/events/${eventID}`);
+  };
+
+  const handleComplete = () => {
+    // TODO: Change User ID to the current user
+
+    if (mode === 'edit') {
+      editCCP({
+        variables: {
+          id: ccpID,
+          name: ccpName,
+          eventId: eventID,
+        },
+      });
+    } else {
+      addCCP({
+        variables: {
+          name: ccpName,
+          eventId: eventID,
+          createdBy: 4,
+        },
+      });
+    }
+
+    // TODO: Check for valid eventID
+
+    history.replace(`/events/${eventID}`);
+  };
+
+  const content = (
+    <ValidatorForm onSubmit={handleComplete}>
+      <FormField
+        label="Name:"
+        placeholder="Create A Name"
+        isValidated
+        onChange={handleNameChange}
+        value={ccpName}
+      />
+      <FormField
+        label="Event Location:"
+        placeholder="Enter Location Here"
+        isValidated
+        onChange={handleLocationChange}
+        value={eventLocation}
+      />
+      <Map />
+      <div className={classes.ccpCompleteDiv}>
+        <CompleteButton disabled={ccpName === ''} buttonText="Complete" />
+      </div>
+    </ValidatorForm>
+  );
+
+  return (
+    <div className={classes.ccpWrapper}>
+      <div className={classes.ccpFormTopSection}>
+        <div className={classes.ccpHeader}>
+          <Typography variant="h3">
+            {mode === 'new' ? 'Create a CCP' : 'Edit CCP'}
+          </Typography>
+          <div className={classes.ccpBtnPosition}>
+            <Button
+              color="primary"
+              className={classes.ccpCancelBtn}
+              onClick={handleCancel}
+            >
+              Cancel
+            </Button>
+          </div>
+        </div>
+      </div>
+      <div className={classes.ccpForm}>{content}</div>
+    </div>
+  );
+};
+
+export default CCPFormPage;
--- a/src/components/CCPForm/CCPFormPage.tsx
+++ b/src/components/CCPForm/CCPFormPage.tsx
@@ -11,7 +11,6 @@ import CompleteButton from './CompleteButton';
 import Map from '../EventCreationPage/Map';
 import FormField from '../common/FormField';
 import {
-  GET_ALL_CCPS,
   GET_CCP_BY_ID,
   GET_CCPS_BY_EVENT_ID,
   CCP,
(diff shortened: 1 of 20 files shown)


src/components/EventDashboard/CCPTabPanel.tsx
```diff
@@ -105,19 +109,76 @@ const EnhancedTableHead = (props: EnhancedTableProps) => {
...
+  };
+
+  //  Writing to cache when deleting ccp
+  const [deleteCCP] = useMutation(DELETE_CCP, {
+    update(cache, { data: { deleteCollectionPoint } }) {
+      if (!deleteCollectionPoint) {
+        return;
+      }
+      let { collectionPointsByEvent } = cache.readQuery<any>({
+        query: GET_CCPS_BY_EVENT_ID,
```
wongcmichelle, line 157: If/when we add a `GET_ALL_CCPS` query, we should probably update the cache for that query here too.

src/components/CCPForm/CCPFormPage.tsx
```diff
@@ -112,24 +134,24 @@ const CCPFormPage = ({
...
+          eventId: eventId,
         },
       });
     } else {
       addCCP({
         variables: {
           name: ccpName,
-          eventId: eventID,
+          eventId: eventId,
           createdBy: 4,
```
wongcmichelle: Just want to note, if you get an error when creating a new CCP, it's likely that the `createdBy` id is invalid. Update it to a valid id then try again :)

src/components/EventDashboard/CCPTabPanel.tsx
```diff
@@ -55,12 +61,6 @@ const useStyles = makeStyles({
 });
 
 const useOptions = makeStyles({
-  root: {
```
wongcmichelle, line 58: This was unused / opting to rely on the styles defined in `OptionPopper`

src/components/EventDashboard/CCPTabPanel.tsx
```diff
@@ -220,13 +287,28 @@ const CCPTabPanel = ({ eventId }: { eventId: string }) => {
...
+          open={openConfirmDelete}
+          title="Are you sure you want to delete the following CCP?"
+          body={
+            <>
+              <span style={{ color: Colours.Secondary, fontWeight: 500 }}>
+                {(selectedCCP as CCP).name}
+              </span>
+              {' cannot be recovered when deleted.'}
+            </>
+          }
```
petorsong, line 281: looking at the other uses of `ConfirmModal`, should we be switching to this pattern for consistency in both code and styling?

petorsong (commented):
- when the modal pops up to delete a CCP, clicking anywhere except for the red `delete` button causes the modal to go away - can you confirm with design that this is expected behaviour? i would've thought that if we have a cancel button, clicking anywhere but the two buttons shouldn't do anything
- out of curiosity, is there a good reason why we have `GET_ALL_CCPS` as a query? we don't have a usecase where we need all CCPs at once and the only good reason i can think of is that it would let us load all of them into cache at once and interact with that easily while filtering out the ones we want per event when we need them
  - but we also have `GET_CCPS_BY_EVENT_ID` and we update that anyways? seems somewhat redundant

src/components/EventDashboard/CCPTabPanel.tsx
wongcmichelle: There's no need to switch to this pattern. This was necessary here because part of the "body" text of the modal/dialog had to be styled differently according to designs.

wongcmichelle:
> 1. when the modal pops up to delete a CCP, clicking anywhere except for the red `delete` button causes the modal to go away - can you confirm with design that this is expected behaviour? i would've thought that if we have a cancel button, clicking anywhere but the two buttons shouldn't do anything
> 2. out of curiosity, is there a good reason why we have `GET_ALL_CCPS` as a query? we don't have a usecase where we need all CCPs at once and the only good reason i can think of is that it would let us load all of them into cache at once and interact with that easily while filtering out the ones we want per event when we need them
>   
>   * but we also have `GET_CCPS_BY_EVENT_ID` and we update that anyways? seems somewhat redundant

1. So I noticed the behaviour is the same in CCP dashboard with the patient detail dialog (clicking anywhere closes the dialog). I can ask the designers.
2. `GET_ALL_CCPS` was added with the new/edit CCP flow. The only use case I can think of for it is if we want unique event names reinforced by the front end. I'm happy to update our codebase and remove uses of `GET_ALL_CCPS` from `CCPTabPanel.tsx` and `CCPFormPage.tsx` (it's not used anywhere else) if that's what we decide to do.

petorsong:
> 
> 1. So I noticed the behaviour is the same in CCP dashboard with the patient detail dialog (clicking anywhere closes the dialog). I can ask the designers.
> 2. `GET_ALL_CCPS` was added with the new/edit CCP flow. The only use case I can think of for it is if we want unique event names reinforced by the front end. I'm happy to update our codebase and remove uses of `GET_ALL_CCPS` from `CCPTabPanel.tsx` and `CCPFormPage.tsx` (it's not used anywhere else) if that's what we decide to do.

1. That would be great, I would prefer/expect that all dialogs would have the same UX
2. Let's nix `GET_ALL_CCPS`, no reason to have code that we're not planning to use; if we do want unique CCP names it would require additional refactoring anyways

wongcmichelle:
> > 1. So I noticed the behaviour is the same in CCP dashboard with the patient detail dialog (clicking anywhere closes the dialog). I can ask the designers.
> > 2. `GET_ALL_CCPS` was added with the new/edit CCP flow. The only use case I can think of for it is if we want unique event names reinforced by the front end. I'm happy to update our codebase and remove uses of `GET_ALL_CCPS` from `CCPTabPanel.tsx` and `CCPFormPage.tsx` (it's not used anywhere else) if that's what we decide to do.
> 
> 1. That would be great, I would prefer/expect that all dialogs would have the same UX
> 2. Let's nix `GET_ALL_CCPS`, no reason to have code that we're not planning to use; if we do want unique CCP names it would require additional refactoring anyways

1. Can we leave additional changes out of this PR and do a more comprehensive refactor of modals/dialogs? There's some other stuff like naming... sometimes we call them modals and sometimes we call them dialogs (depending on the component).
2. Will do 👍

petorsong:
here's the new ticket for cleaning up and standardizing modal behaviour
https://www.notion.so/uwblueprintexecs/dialog-modal-cleanup-4e0ec85b6e2f41ef8829c53a8a0fa134

src/components/CCPForm/CCPFormPage.tsx
petorsong: let's actually change this to 1 so that staging isn't broken (we only have a single user on staging right now)
```suggestion
          createdBy: 1,
```

src/components/CCPForm/CCPFormPage.tsx
```diff
@@ -84,12 +89,29 @@ const CCPFormPage = ({
 
   const [addCCP] = useMutation(ADD_CCP, {
     update(cache, { data: { addCollectionPoint } }) {
```
aaronabraham311, line 92: I'm a little unclear why there are three queries for adding a CCP. Could you please elaborate on this?

src/components/EventDashboard/CCPTabPanel.tsx
```diff
@@ -120,49 +120,117 @@ const EnhancedTableHead = (props: EnhancedTableProps) => {
...
+      });
+
+      // Update GET_ALL_CCPS
+      const { collectionPoints } = cache.readQuery<any>({
+        query: GET_ALL_CCPS,
+      });
+
+      if (collectionPoints) {
+        updatedCCPsList = collectionPoints.filter(
+          (ccp) => ccp.id !== (selectedCCP as CCP).id
```
aaronabraham311: Just curious, why did you do this: `(selectedCCP as CCP).id`?

aaronabraham311 (commented):
Great work! Just a few comments for my own edification.

aaronabraham311:
I don't know whether this is expected behaviour or not but when I create a CCP with a name and location and then edit it, the location does not appear but the name does. It seems strange that the two fields have different behaviours. 
Screen Shot 2020-08-28 at 2 16 41 PM

petorsong:
> 
> 
> I don't know whether this is expected behaviour or not but when I create a CCP with a name and location and then edit it, the location does not appear but the name does. It seems strange that the two fields have different behaviours.
> Screen Shot 2020-08-28 at 2 16 41 PM

@aaronabraham311 this is due to location data not being implemented yet - on the frontend we don't store anything for that field anyways so attempting to edit it will come back empty

petorsong:
@wongcmichelle https://github.com/uwblueprint/paramedics-react/pull/29 merged in many of the camel case `eventID` -> `eventId` changes, so you'll have to resolve the merge conflicts

src/components/EventDashboard/CCPTabPanel.tsx
wongcmichelle: This is because on line 129: `const [selectedCCP, setSelectedCCP] = React.useState<CCP | null>(null);`, `selectedCCP` is defined as `CCP | null` and `null` has no property `id`. But we know `selectedCCP` cannot be null when we reach this point in the code, so we're casting it as type `CCP`.

src/components/CCPForm/CCPFormPage.tsx
wongcmichelle: I'm not sure if this is the clarification you need, but `cache.readQuery` reads from the cache, while `useQuery` actually attempts to fetch data from the backend if it's not found in the cache.

petorsong (approved):

src/components/CCPForm/CCPFormPage.tsx
```diff
@@ -51,26 +55,28 @@ const useStyles = makeStyles({
...
 
-  const { data } = useQuery(mode === 'new' ? GET_ALL_CCPS : GET_CCP_BY_ID, {
-    variables: { id: ccpID },
-  });
+  const { data } = useQuery(
+    mode === 'edit' ? GET_CCP_BY_ID : GET_CCPS_BY_EVENT_ID,
+    {
+      variables: { id: ccpId, eventId },
+    }
+  );
```
wongcmichelle, line 78: Technically, we don't need to run the `GET_CCPS_BY_EVENT_ID` query, but I added it here for two reasons:
1. We can't conditionally call hooks, so we have to pass something valid to `useQuery` (for both mode new and edit)
2. In `addCCP`, we call `cache.readQuery` to get `collectionPointsByEvent` and add the new CCP to the cache. `readQuery` throws an error if the data isn't available in the cache and never attempts to fetch data from the server (see [here](https://www.apollographql.com/docs/react/caching/cache-interaction/#readquery)). Calling `useQuery` with `GET_CCPS_BY_EVENT_ID` here basically guarantees that the data we need will be in the cache (and the error won't be thrown by `readQuery`).

petorsong: thanks for following up, it's nice that the `readQuery` call is not as expensive as going to the server

src/components/EventDashboard/CCPTabPanel.tsx
aaronabraham311: Got it thanks!

aaronabraham311 (approved):

PR #36, opened 2020-09-03
Patient delete button

aaronabraham311:
[Notion task](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=e0152f69c6d1415f8892686284db86db)

## Changes
- Created delete button according to designs
- Added GraphQL mutation for deleting patients for frontend

## Testing
- Navigate to a particular patient in your frontend (`/events/eventId/ccps/ccpId/patients/patientId`)
- Ensure that the delete patient button is present and the confirm modal is working as expected
- When you click 'delete', check the backend and determine whether the patient in question has a `status: Deleted`
- Make sure that the CCP dashboard is updated with the new deleted status when you navigate back
## Screenshots
Kapture 2020-09-03 at 13 51 26


src/containers/PatientProfilePage.tsx
```diff
@@ -154,6 +162,25 @@ const PatientProfilePage = ({
...
+
+  const handldeDeleteConfirm = () => {
+    if (mode === 'edit') {
+      deletePatient({
+        variables: {
+          id: patientId,
+        },
+      });
+    }
+    history.replace('/');
```
petorsong: how come we're redirecting to the home page here? also would `history.push` be more appropriate?

src/styles/Constants.ts
```diff
@@ -2,8 +2,8 @@ export enum Colours {
   Primary = '#05162F',
   Secondary = '#2E5584',
   SecondaryHover = '#284D7A',
-  Danger = '#AC3434',
-  DangerHover = '#9B2F2F',
+  Danger = '#B12323',
+  DangerHover = '#DD2B2B',
```
petorsong, line 6: what's prompting this change? do we have usages of either of these constants that are have different colours in the mocks?

petorsong (changes requested):
should we also remove `DELETED` as one of the options for statuses?
image

src/containers/PatientProfilePage.tsx
aaronabraham311: I'm not totally sure where we should be redirecting to. As for using `history.push`: it doesn't make sense to have the deleted patient in our history. Ideally, the user should not have a method for coming back to that patient.

src/styles/Constants.ts
aaronabraham311: Charmaine made some changes to the colours, so I updated them accordingly.

aaronabraham311:
> should we also remove DELETED as one of the options for statuses?

I will be addressing that in the next PR. This one is only for the patient deletion button.

src/containers/PatientProfilePage.tsx
petorsong: let's redirect back to the CCP dashboard for now, we can revisit this if product mentions otherwise

src/components/PatientProfile/PatientProfilePage.tsx
```diff
@@ -97,8 +103,26 @@ const PatientProfilePage = ({
     },
   });
   const [editPatient] = useMutation(EDIT_PATIENT);
+  const [deletePatient] = useMutation(DELETE_PATIENT, {
```
aaronabraham311: I am really not sure if this is best practice with updating the cache. I had to implement a cache update because using `const [deletePatient] = useMutation(DELETE_PATIENT)` was not updating the cache by itself (unlike `editPatient` which is weird).

src/styles/Constants.ts
```diff
@@ -2,8 +2,8 @@ export enum Colours {
   Primary = '#05162F',
   Secondary = '#2E5584',
   SecondaryHover = '#284D7A',
-  Danger = '#AC3434',
-  DangerHover = '#9B2F2F',
+  Danger = '#B12323',
```
aaronabraham311, line 5: Updated due to design change

src/components/PatientProfile/PatientProfilePage.tsx
petorsong: this may be related to the fact that the delete mutation isn't returning the ID as of now? cc @wongcmichelle

src/components/PatientProfile/PatientProfilePage.tsx
```diff
@@ -153,6 +177,25 @@ const PatientProfilePage = ({
...
 
+  const handleDeleteClick = () => {
+    setDeleteClicked(true);
+  };
+
+  const handleDeleteCancel = () => {
+    setDeleteClicked(false);
+  };
+
+  const handleDeleteConfirm = async () => {
```
petorsong: is there a reason this is `async`?

src/components/PatientProfile/PatientProfilePage.tsx
```diff
@@ -433,6 +476,19 @@ const PatientProfilePage = ({
             </>
           )}
           <CompletePatientButton />
+          {mode === 'edit' && (
+            <DeletePatientButton handleClick={handleDeleteClick} />
+          )}
+          {mode === 'edit' && (
```
petorsong: can these 2 `mode === 'edit'` conditions be combined?

petorsong (changes requested):

src/components/PatientProfile/PatientProfilePage.tsx
aaronabraham311: I updated my backend too with Michelle's changes so this should not be an issue anymore

petorsong (changes requested):
getting another infinite render loop error again - likely has to do with `useEffect`
```
Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
    in ForwardRef(InputBase) (created by WithStyles(ForwardRef(InputBase)))
    in WithStyles(ForwardRef(InputBase)) (created by ForwardRef(Input))
    in ForwardRef(Input) (created by WithStyles(ForwardRef(Input)))
    in WithStyles(ForwardRef(Input)) (created by ForwardRef(TextField))
    in div (created by ForwardRef(FormControl))
    in ForwardRef(FormControl) (created by WithStyles(ForwardRef(FormControl)))
    in WithStyles(ForwardRef(FormControl)) (created by ForwardRef(TextField))
    in ForwardRef(TextField) (created by WithStyles(ForwardRef(TextField)))
    in WithStyles(ForwardRef(TextField)) (created by TextValidator)
    in TextValidator (at FormField.tsx:72)
    in FormField (at PatientProfilePage.tsx:371)
    in form (created by ValidatorForm)
    in ValidatorForm (at PatientProfilePage.tsx:367)
    in div (created by Styled(MuiBox))
    in Styled(MuiBox) (at PatientProfilePage.tsx:366)
    in div (created by Styled(MuiBox))
    in Styled(MuiBox) (at PatientProfilePage.tsx:309)
    in PatientProfilePage (at App.tsx:94)
    in component (created by Context.Consumer)
    in Route (at App.tsx:91)
    in Switch (at App.tsx:25)
    in SnackbarProvider (at App.tsx:22)
    in ThemeProvider (at App.tsx:21)
    in App (at src/index.tsx:16)
    in ApolloProvider (at src/index.tsx:15)
    in Router (created by BrowserRouter)
    in BrowserRouter (at src/index.tsx:14)
```

Paramedics React - Mozilla Firefox 2020-10-01 21-11-48

aaronabraham311:
This is what I was concerned about with addressing the `useEffect()` warnings. Seems like I added the `formFields` dependency on `useEffect()` which actually sets the the `formFields` states. I have a fix which requires using functional `setState()`; will get that pushed tomorrow.

aaronabraham311:
There is an additional `setState()` issue that I will fix in the cleanup ticket

petorsong (dismissed):
🚀

wongcmichelle:
It looks like the cache is **updated incorrectly only when** I go directly to `/events/eventId/ccps/ccpId/patients/patientId` (without entering the form through the CCP Dashboard). 

The cache was wrong for both the patient by id and patients query:
image

@petorsong Not sure if this needs to be fixed? (It probably should be)

The cache is consistently correctly updated when I delete a patient from the popover:
image

_Just a side thought:_ Perhaps we should hide the popover or at least the delete patient option when a user is already deleted. (Same goes for in the edit patient form: we might want to hide the "DELETE PATIENT" button)

src/components/PatientProfile/PatientProfilePage.tsx
wongcmichelle: ~~Delete not updating the cache automatically is expected as explained here: https://uwblueprint.slack.com/archives/G0149QDHPEV/p1600823248003100~~

Disregard this comment because I forgot that our delete is updating the status and not actually deleting the entity :)

src/components/PatientProfile/PatientProfilePage.tsx
```diff
@@ -131,16 +155,15 @@ const PatientProfilePage = ({
...
-        notes,
-        status,
-        runNumber,
-      });
+        gender: gender,
+        age: age,
+        notes: notes,
+        status: status,
+        runNumber: runNumber,
+      }));
```
wongcmichelle: Out of curiosity, why did you make these updates? Should make no difference because we were just using the [shorthand](https://alligator.io/js/object-property-shorthand-es6/) before.

wongcmichelle: Just checked and it actually makes our linter unhappy :(

src/components/PatientProfile/PatientProfilePage.tsx
```diff
@@ -149,10 +172,29 @@ const PatientProfilePage = ({
 
   useEffect(() => {
     if (mode === 'new' && barcodeValue) {
-      setFormFields({ ...formFields, barcodeValue });
+      setFormFields({ ...formFields, barcodeValue: barcodeValue });
```
wongcmichelle: Same comment here about shorthand

wongcmichelle (commented):
2 small things:

1. The linter should fix the shorthand issue I mentioned
2. The delete button in the confirm dialog doesn't match Figma:
image

Otherwise LGTM, thanks :)

petorsong:
> It looks like the cache is **updated incorrectly only when** I go directly to `/events/eventId/ccps/ccpId/patients/patientId` (without entering the form through the CCP Dashboard).
> 
> The cache was wrong for both the patient by id and patients query:
> ...
> 
> @petorsong Not sure if this needs to be fixed? (It probably should be)

this is a good catch - please look into this as part of this PR @aaronabraham311 

> 
> The cache is consistently correctly updated when I delete a patient from the popover:
> ...
> 
> _Just a side thought:_ Perhaps we should hide the popover or at least the delete patient option when a user is already deleted. (Same goes for in the edit patient form: we might want to hide the "DELETE PATIENT" button)

this popover thing will be handled in a separate ticket
https://www.notion.so/uwblueprintexecs/As-a-user-I-want-the-flexibility-to-delete-a-patient-entry-f61270edfb854f21a4d1302a247933de

src/components/PatientProfile/PatientProfilePage.tsx
aaronabraham311: Ok let me walk through my logic on why I made this change:
1. We want to get rid of the warnings with `useEffect()` complaining about unneeded dependencies. We can either add the dependency or we can refactor
2. Adding the dependency leads to infinite loops. Specifically, this warning is complaining that `formFields` should be a dependency but the hook itself changes `formFields`, meaning that the hook gets called again and again.
3. To refactor, we need to remove all instances of `formFields`. I thought the best way of doing this is the functional `setFormFields` that I wrote
4. It seems to have passed our linter. I did notice that the linter complains but linting it causes no issues.

petorsong: i think michelle is referring specifically to the prop assignment shorthand (i.e. using `gender: gender` when `gender` is enough)

can either of you explain what you mean by the linter complaining? it seems fine on my end (e.g. no errors or warnings show up)

aaronabraham311: I get red lines on VS Code regarding property shorthands from eslint

petorsong: @aaronabraham311 do you have issues with the shorthand syntax? if vscode is complaining, i'd prefer to stick with the shorthand if possible

petorsong: i realize that the linter actually fixes this syntax and makes it into the shorthand, so let's revert this so it just goes with the original shorthand

aaronabraham311:
Blocked by https://github.com/uwblueprint/paramedics-web/pull/68.

src/components/PatientProfile/PatientProfilePage.tsx
```diff
@@ -435,7 +418,7 @@ const PatientProfilePage = ({
                 notes: (e.target as HTMLInputElement).value,
               });
             }}
-            value={formFields.notes}
+            value={formFields.notes ? formFields.notes : ''}
```
petorsong: nit: this can also be simplified to the following - either option is fine and is a stylistic choice, so up to you if you want to make this change
```suggestion
            value={formFields.notes || ''}
```

petorsong (dismissed):
looks gucci ✨

src/graphql/mutations/patients.ts
```diff
@@ -124,6 +124,15 @@ export const EDIT_PATIENT = gql`
...
+    deletePatient(id: $id) {
+      id
+      collectionPointId {
+        id
+        eventId {
+          id
+        }
+      }
+      status
+    }
```
wongcmichelle:
```suggestion
    deletePatient(id: $id) {
      id
      gender
      age
      runNumber
      barcodeValue
      collectionPointId {
        id
        name
        eventId {
          id
          name
          eventDate
        }
      }
      triageLevel
      status
      notes
      updatedAt
      transportTime
      hospitalId {
        id
        name
      }
      ambulanceId {
        id
        vehicleNumber
      }
    }
```

wongcmichelle (dismissed):
Awesome, everything is working! thanks :) 

It's probably best to return everything in the `deletePatient` mutation though, just to mitigate any potential issues with mismatching fields in the cache.

wongcmichelle (approved):

petorsong (approved):

PR #40, opened 2020-10-07
Handle "Enter" key in manual barcode

patrick-du:
[Notion ticket](https://www.notion.so/uwblueprintexecs/Enter-Barcode-Page-pressing-the-Enter-key-does-not-pass-input-6c7eacc467fe4d73915e497ae63189d2)

## Changes
- ~~Updated `FormField.tsx` with a new `onKeyPress` prop~~
- ~~Reason for prop is because `TextField`, the component used in `FormField.tsx`, doesn't seem to have this as a prop~~ 
- ~~Additionally, onChange does not handle `React.KeyboardEvent`~~
- ~~On "Enter" key press, directs the user to the next step of the flow~~
- **After CR suggestions**
    - Use `onSubmit` to handle "Enter" key
## Testing
- Type in a barcode manually and press the "Enter" key on your keyboard
- This should redirect you to the next barcode page with your input from the previous page in the barcode field 
- If there was no input, you will stay on the same page

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [x] make sure build passes

### DIFF
--- a/src/components/EnterBarcode/EnterBarcodePage.tsx
+++ b/src/components/EnterBarcode/EnterBarcodePage.tsx
@@ -49,6 +49,10 @@ const EnterBarcodePage = ({
     }
   };
 
+  const handleFormSubmit = (e) => {
+    return barcode !== '' ? handleEnterBarcode() : e.preventDefault();
+  };
+
   return (
     <Box minHeight="100vh">
       <Box
@@ -67,11 +71,7 @@ const EnterBarcodePage = ({
         </Button>
       </Box>
       <Box padding="56px">
-        <form
-          onSubmit={(e) =>
-            barcode !== '' ? handleEnterBarcode() : e.preventDefault()
-          }
-        >
+        <form onSubmit={handleFormSubmit}>
           <FormField
             label="Barcode:"
             placeholder="Enter barcode here"
--- a/src/components/EnterBarcode/EnterBarcodePage.tsx
+++ b/src/components/EnterBarcode/EnterBarcodePage.tsx
@@ -21,35 +21,31 @@ const EnterBarcodePage = ({
   const [barcode, setBarcode] = useState<string>('');
 
   const handleEnterBarcode = () => {
-    if (barcode !== '') {
-      if (!loading && data) {
-        const selectedPatient = data.patients.filter(
-          (patient) => patient.barcodeValue === barcode
+    if (!loading && data) {
+      const selectedPatient = data.patients.filter(
+        (patient) => patient.barcodeValue === barcode
+      );
+      if (selectedPatient.length > 0) {
+        // Found patient
+        const {
+          collectionPointId: {
+            id: patientCcpId,
+            eventId: { id: patientEventId },
+          },
+          id,
+        } = selectedPatient[0];
+        // Redirect to patient profile
+        history.push(
+          `/events/${patientEventId}/ccps/${patientCcpId}/patients/${id}`
         );
-        if (selectedPatient.length > 0) {
-          // Found patient
-          const {
-            collectionPointId: {
-              id: patientCcpId,
-              eventId: { id: patientEventId },
-            },
-            id,
-          } = selectedPatient[0];
-          // Redirect to patient profile
-          history.push(
-            `/events/${patientEventId}/ccps/${patientCcpId}/patients/${id}`
-          );
-        } else {
-          // No existing patient
-          history.push(
-            `/events/${eventId}/ccps/${ccpId}/patients/new/${barcode}`
-          );
-        }
       } else {
+        // No existing patient
         history.push(
           `/events/${eventId}/ccps/${ccpId}/patients/new/${barcode}`
         );
       }
+    } else {
+      history.push(`/events/${eventId}/ccps/${ccpId}/patients/new/${barcode}`);
     }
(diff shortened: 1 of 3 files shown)


petorsong:
`onKeyPress` is an action that is no longer recommended as it has been removed from web standards (which is why material UI may have chosen to not support it) - [see this doc for more info](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onkeypress)

instead, i would prefer to leverage an action like `onSubmit` instead so we're not explicitly listening for a specific key press (i.e. enter) and instead just submitting the form normally, as if we had clicked the submit button

petorsong (approved):
real straightforward 🥇

JasonYG (dismissed):
Nice and simple change :)

petorsong (changes requested):
see new comments on notion ticket for additional changes necessary
https://www.notion.so/uwblueprintexecs/Enter-Barcode-Page-pressing-the-Enter-key-does-not-pass-input-6c7eacc467fe4d73915e497ae63189d2

rickymao:
LGTM!

rickymao (dismissed):

patrick-du:
For the additional changes, I added a disabled check for the button for the visual effect, and for users that press the "Enter" key, I added an additional check in the method that handles the button, `handleEnterBarcode`.

rickymao (dismissed):
Looks Spooky To Me 👻

src/components/EnterBarcode/EnterBarcodePage.tsx
```diff
@@ -21,31 +21,35 @@ const EnterBarcodePage = ({
...
-          collectionPointId: {
-            id: patientCcpId,
-            eventId: { id: patientEventId },
-          },
-          id,
-        } = selectedPatient[0];
-        // Redirect to patient profile
-        history.push(
-          `/events/${patientEventId}/ccps/${patientCcpId}/patients/${id}`
+    if (barcode !== '') {
```
petorsong: this check doesn't seem necessary - if we only call `handleEnterBarcode` when the form is submitted (i.e. when `barcode` is not an empty string), then this condition is always true as we cannot submit the form when `barcode` is an empty string

petorsong (changes requested):
i'm seeing some weird behaviour when i hit enter in the input field while it's empty - it basically reloads the page but appends `?` to the route; ideally we don't do anything when trying to hit enter with an empty barcode, but if this proves difficult we can keep this current behaviour (not ideal though)

src/components/EnterBarcode/EnterBarcodePage.tsx
patrick-du: Sounds good - in the new commit, I performed the check within `onSubmit` rather than inside of `handleEnterBarcode` since this function should only be routing the user rather than worrying about empty strings

patrick-du: Also, I prevented the reloading with `e.preventDefault()` - hopefully that's the correct way to do so

src/components/EnterBarcode/EnterBarcodePage.tsx
```diff
@@ -67,7 +67,11 @@ const EnterBarcodePage = ({
         </Button>
       </Box>
       <Box padding="56px">
-        <form>
+        <form
+          onSubmit={(e) =>
+            barcode !== '' ? handleEnterBarcode() : e.preventDefault()
```
patrick-du: Is this alright or should it be put in a function and called onSubmit instead?
```
const handleFormSubmission = (e) => {
     barcode !== '' ? handleEnterBarcode() : e.preventDefault()
}

...

<form onSubmit={(e) => handleFormSubmission(e)}>
```

petorsong: i think this is just a matter of preference, up to you if you want to make the change - it can be cleaner and easier to edit in the future

also small nit for your example - `<form onSubmit={handleFormSubmit}>` should work as well - if you are gonna make the change i would test it first though

petorsong (dismissed):
:shipit:

src/components/EnterBarcode/EnterBarcodePage.tsx
patrick-du: Ok I've updated it - this ticket should be complete 😅

petorsong (approved):
🥇

src/components/EnterBarcode/EnterBarcodePage.tsx
```diff
@@ -49,6 +49,10 @@ const EnterBarcodePage = ({
     }
   };
 
+  const handleFormSubmit = (e) => {
+    return barcode !== '' ? handleEnterBarcode() : e.preventDefault();
```
JasonYG, line 53: just wanted to confirm, we have `e.preventDefault()` so the form doesn't submit if the barcode is empty, right?

JasonYG (approved):

JasonYG:
Changes look good!!

rickymao (approved):
Looks spooky to me!

PR #42, opened 2020-10-13
Archive Event Cleanup

carelynntsai:
[Notion ticket](https://www.notion.so/uwblueprintexecs/archive-event-card-cleanup-1b64ce8fa69a45fd93020db66d2ece6e)

## Changes
- Highlights newly archived/unarchived events
- After archiving/unarchiving event, routes patient to corresponding tab

## Screenshots
image
image

## Testing
- Try archiving/unarchiving event, check you are taken to the correct tab and the event is highlighted in blue
- Try navigating away, then back from the page and check event is no longer highlighted

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [ ] make sure build passes
- [ ] ping `@ps` in `#paramedics-dev`

### DIFF
--- a/src/components/EventOverview/EventsPage.tsx
+++ b/src/components/EventOverview/EventsPage.tsx
@@ -42,7 +42,7 @@ const EventsPage = () => {
   const classes = useStyles();
   const location = useLocation<LocationState>();
   const { addedEventId } = location.state || { addedEventId: null };
-  const [archivedEventId, setArchivedEventId] = useState('');
+  const [lastUpdatedEventId, setLastUpdatedEventId] = useState(addedEventId);
   const [selectedTab, setTab] = useState(0);
   const [eventToBeDeleted, setEventToBeDeleted] = useState<Event | null>(null);
 
@@ -59,6 +59,7 @@ const EventsPage = () => {
     event: React.ChangeEvent<unknown>,
     newValue: number
   ) => {
+    setLastUpdatedEventId(null);
     setTab(newValue);
   };
 
@@ -99,7 +100,7 @@ const EventsPage = () => {
         isActive: false,
       },
     });
-    setArchivedEventId(event.id);
+    setLastUpdatedEventId(event.id);
     setTab(1);
   };
 
@@ -113,7 +114,7 @@ const EventsPage = () => {
         isActive: true,
       },
     });
-    setArchivedEventId(event.id);
+    setLastUpdatedEventId(event.id);
     setTab(0);
   };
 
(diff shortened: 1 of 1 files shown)


rickymao:
Is it supposed to be still highlighted if you switch from 'Archived Events' to 'Current Events'.

rickymao (commented):
Looks Spooky To Me 👻

src/components/EventOverview/EventsPage.tsx
```diff
@@ -42,6 +42,7 @@ const EventsPage = () => {
   const classes = useStyles();
   const location = useLocation<LocationState>();
   const { addedEventId } = location.state || { addedEventId: null };
+  const [archivedEventId, setArchivedEventId] = useState('');
```
petorsong: nit? - is there a reason we use empty string instead of `null` here?

src/components/EventOverview/EventsPage.tsx
```diff
@@ -149,7 +155,9 @@ const EventsPage = () => {
                 date={event.eventDate}
                 eventTitle={event.name}
                 isActive={event.isActive}
-                isNew={event.id === addedEventId}
+                isNew={
+                  event.id === addedEventId || event.id === archivedEventId
```
petorsong: there is a scenario where if you add a new event and then immediately archive a **different** event, you get 2 cards that are highlighted at once - is it a big deal to get it so that only the last event that changed gets highlighted?

petorsong (dismissed):
looks fine, see comments for potential changes

rickymao (dismissed):

src/components/EventOverview/EventsPage.tsx
carelynntsai: typescript doesn't let you use null, because I set it to whatever the new archived event's id on this page which a type string

carelynntsai:
Only the last updated event should be highlighted now. Also if you switch tabs the event card won't be highlighted anymore.

rickymao (approved):
Looks spooky to me! Weird maybe unexpected behaviour, if you select the same tab it removes the highlight. Not sure if it is a big deal or not, but I'll flag it here. Good job overall!

petorsong (approved):
unhighlighting on selecting the same tab might actually be desired behaviour, but it's only a minor issue in the worst case scenario - looks good otherwise

PR #45, opened 2020-10-16
Update user roles

sarafong:
[Notion ticket](https://www.notion.so/uwblueprintexecs/update-user-access-levels-1488ff68ce2a419a85c82d2f690b21d2)

[Respective backend change](https://github.com/uwblueprint/paramedics-web/pull/70)

## Changes
- Updated `AccessLevelSelector` to use new ID values for roles instead of access level enum 
- Deleted `AccessLevel` enum and used role ID as value in radio group instead
- Updated GraphQL to use `roleId` instead of `accessLevel`
- Ensure backend change is merged at the same time

## Testing
1. To test, you must be on `update-role-names` branch in `paramedics-web` and must have run the migration
2. Add or edit team member in Resource Management and ensure you can toggle between roles
3. Ensure submission is successful and verify that the new role ID is correct in GraphQL playground

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [ ] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [ ] make sure build passes

### DIFF
--- a/src/components/ResourceForm/AccessLevelSelector.tsx
+++ b/src/components/ResourceForm/AccessLevelSelector.tsx
@@ -45,7 +45,7 @@ const AccessLevelSelector = ({
   currentValue,
   handleChange,
 }: {
-  currentValue: string;
+  currentValue: number;
   handleChange: (e: any) => any;
 }) => {
   const classes = useRadioStyles();
--- a/src/components/ResourceForm/AccessLevelSelector.tsx
+++ b/src/components/ResourceForm/AccessLevelSelector.tsx
@@ -5,7 +5,6 @@ import FormControlLabel from '@material-ui/core/FormControlLabel';
 import RadioGroup from '@material-ui/core/RadioGroup';
 import Radio from '@material-ui/core/Radio';
 import Typography from '@material-ui/core/Typography';
-import { AccessLevel } from '../../graphql/queries/users';
 
 const useRadioStyles = makeStyles({
   root: {
@@ -51,9 +50,9 @@ const AccessLevelSelector = ({
 }) => {
   const classes = useRadioStyles();
   const accessLevels = [
-    { val: AccessLevel.COMMANDER, label: 'Commander', roleId: 1 },
-    { val: AccessLevel.SUPERVISOR, label: 'Supervisor', roleId: 2 },
-    { val: AccessLevel.DISPATCH, label: 'Dispatch', roleId: 3 },
+    { label: 'Commander', roleId: 1 },
+    { label: 'Supervisor', roleId: 2 },
+    { label: 'Dispatch', roleId: 3 },
   ];
   return (
     <Container className={classes.root}>
--- a/src/components/ResourceForm/AccessLevelSelector.tsx
+++ b/src/components/ResourceForm/AccessLevelSelector.tsx
@@ -51,9 +51,9 @@ const AccessLevelSelector = ({
 }) => {
   const classes = useRadioStyles();
   const accessLevels = [
-    { val: AccessLevel.ADMIN, label: 'Admin' },
-    { val: AccessLevel.SUPERVISOR, label: 'CCP Supervisor' },
-    { val: AccessLevel.COMMANDER, label: 'Dispatch Coordinator' },
+    { val: AccessLevel.COMMANDER, label: 'Commander', roleId: 1 },
+    { val: AccessLevel.SUPERVISOR, label: 'Supervisor', roleId: 2 },
+    { val: AccessLevel.DISPATCH, label: 'Dispatch', roleId: 3 },
   ];
   return (
     <Container className={classes.root}>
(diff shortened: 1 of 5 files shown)


src/components/ResourceForm/AccessLevelSelector.tsx
```diff
@@ -68,11 +67,11 @@ const AccessLevelSelector = ({
       >
         {accessLevels.map((accessLevel) => (
           <FormControlLabel
-            value={accessLevel.val}
+            value={accessLevel.roleId.toString()}
```
petorsong, line 70: nit: can't the `value` prop simply take a number? is this necessary?

src/components/ResourceForm/UserFormPage.tsx
```diff
@@ -114,7 +113,7 @@ const UserFormPage = ({
   };
 
   const handleRoleChange = (e: any) => {
-    setRole(e.target.value);
+    setRole(Number(e.target.value));
```
petorsong: if `value` is made a number, this wouldn't be necessary

src/components/ResourceForm/UserFormPage.tsx
```diff
@@ -187,7 +186,7 @@ const UserFormPage = ({
             value={email}
           />
           <AccessLevelSelector
-            currentValue={role}
+            currentValue={role.toString()}
```
petorsong: same here

src/graphql/queries/users.ts
```diff
@@ -1,17 +1,11 @@
...
-  ADMIN = 'ADMIN',
-}
-
 export interface User {
   id: string;
   name: string;
   email: string;
   password: string;
-  accessLevel: AccessLevel;
+  roleId: number;
```
petorsong: should this by `string`? see `id` above

petorsong (changes requested):
some confusion around types, but otherwise looks good!

src/components/ResourceForm/UserFormPage.tsx
```diff
@@ -75,22 +74,22 @@ const UserFormPage = ({
 
   const [memberName, setMemberName] = useState<string>('');
   const [email, setEmail] = useState<string>('');
-  const [role, setRole] = useState<AccessLevel>(AccessLevel.SUPERVISOR);
+  const [role, setRole] = useState<number>(2);
```
aaronabraham311: Personally, I think using the role ID numbers are really confusing, because I am not exactly sure what '2' represents unless I look at `AccessLevelSelector.tsx`. That being said, was there a reason why we didn't use an enum to represent the different values in a more readable fashion?

src/graphql/queries/users.ts
```diff
@@ -1,17 +1,11 @@
 import gql from 'graphql-tag';
```
aaronabraham311, line 2: Why can we not just keep the `AccessLevel` enum and define it in terms of numbers instead so that we can use it in conjunction with role ID?

```suggestion
export enum AccessLevel {
    COMMANDER = 1
    SUPERVISOR = 2
    DISPATCH = 3
}
```

aaronabraham311 (commented):
Works as expected! Nice work. I have a few comments below but not sure if you need to make these changes; it's more for readability than functional fixes.

petorsong:
@aaronabraham311 about the enum stuff, it doesn't necessarily need to do be part of this ticket and does overlap with the authentication work

@JasonYG @sarafong can you two coordinate to figure out how this `role`/`accessLevel` business will be handled and who will implement it when?

src/components/ResourceForm/AccessLevelSelector.tsx
sarafong: @petorsong hm I vaguely recall I put this in because it wasn't working with `number`, then I checked the docs and it said `string` was required. but now I just checked again and it says `any` 🤔

will need to verify this when I can test again

src/graphql/queries/users.ts
sarafong: yeah this is actually a good suggestion, I'll try changing it to this. I changed them to be all numbers since I thought it would be more clear what the backend is accepting, but it's probably just more confusing

src/components/ResourceForm/AccessLevelSelector.tsx
sarafong: Just looked into it, this is actually necessary because `RadioGroup`'s value prop casts it to a string by default. Thus the current value doesn't show up as selected unless you also cast this to a string.

src/components/ResourceForm/UserFormPage.tsx
```diff
@@ -14,9 +14,9 @@ import { Colours } from '../../styles/Constants';
 import { ADD_USER, EDIT_USER } from '../../graphql/mutations/users';
 import {
   User,
-  AccessLevel,
   GET_ALL_USERS,
   GET_USER_BY_ID,
+  AccessLevel,
```
petorsong, line 19: nit: this probably didn't need to get moved lol

petorsong (approved):
🌮

aaronabraham311 (approved):

petorsong:
wait how come this PR didn't trigger our CI builds?

src/components/ResourceForm/UserFormPage.tsx
sarafong: oops idk why i did that

PR #46, opened 2020-10-22
Display maps

aaronabraham311:
[Notion ticket](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=419c95b8cc0d4a8ab4b2783744dfb16c)

## Changes
- Created Google Maps page
    - Displays location pins associated with a particular event
    - Built info window component to display pin information when pin is clicked
    - Added green location pin that indicates current location (enable this when prompted)
    - Integrated satellite view
- Added maps navigation to hamburger menu

## Screenshots
Pins displayed
Screen Shot 2020-10-22 at 3 48 49 PM

Info window
Screen Shot 2020-10-22 at 3 54 07 PM

Hamburger menu
Screen Shot 2020-10-22 at 4 04 51 PM
## Testing
Note that pins are supposed to represent the location of CCPs. For testing purposes, the pins displayed in this PR are random pins not associated with any CCP or event.

Before testing:
1. Update your backend: I have added new seeders to put pins in the Waterloo region. This makes testing much easier. Once the backend is updated, you can navigate to the 'Spooky Scary Event' -> click the hamburger menu -> navigate to the maps.
2. Add the Google Map API Key in `paramedics-react.env` which you can get [here](https://www.notion.so/uwblueprintexecs/6a0ad062ce504acc90fa25cb4d2c731f?v=4844a823da724b7ebf2c442a1ee1279f&p=e9959258451a48bba706904335794f57)

Note that this was styled based on the iPad Pro dimensions using the Chrome inspection (I need to define an absolute size of the map frame). Styling is still in progress and depends on the iPads that the paramedics are using.

Here are some tests to run
1. Ensure that all location pins are displayed
2. Ensure that the info window opened when a pin is clicked displays the correct information
3. Ensure that regular maps functionality is present (eg. zoom, current location, satellite vs. road map)

In my experience, the UI to test the changes on the Chrome iPad settings aren't great. If you do have an iPad, open up a tunnel and test it out.

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [ ] make sure build passes


src/components/Maps/MapPage.tsx
```diff
@@ -0,0 +1,136 @@
...
+  },
+}: {
+  match: { params: { eventId: string } };
+}) => {
+  const defaultMap = {
+    center: { lat: 43.470846, lng: -80.538473 },
+    zoom: 11,
+  };
+
+  const getMapOptions = (maps) => {
```
aaronabraham311, line 42: These are the default settings of the map

src/components/Maps/MapPage.tsx
```diff
@@ -0,0 +1,136 @@
...
+    variables: { eventId },
+  });
+
+  const pins: Array<LocationPin> = data && !loading ? data.pinsForEvent : [];
+  const [infoWindowOpen, setInfoWindowOpen] = React.useState(false);
+  const [interestPinTitle, setInterestPinTitle] = React.useState('');
+  const [interestPinLocation, setInterestPinLocation] = React.useState('');
+
+  const initMaps = (map, maps) => {
+    if (!loading) {
```
aaronabraham311: I occasionally run into an issue where the pins don't load when I refresh the page. If you are still running into this issue, let me know. I have tried many fixes and I believe the error is somewhere in this function. Hopefully I have addressed it.

src/components/Maps/MapPage.tsx
```diff
@@ -0,0 +1,136 @@
...
+          map.setCenter(marker.getPosition());
+          setInfoWindowOpen(true);
+          setInterestPinLocation(pin.address);
+          setInterestPinTitle(pin.label);
+        });
+
+        return marker;
+      });
+
+      if (navigator.geolocation) {
```
aaronabraham311: Separate marker for current location (which is why I have a `setIcon` call to distinguish this marker from other markers).

src/components/Maps/InfoWindow.tsx
```diff
@@ -0,0 +1,82 @@
...
+        </Typography>
+      </DialogContent>
+      <DialogActions>
+        <Button size="small" classes={{ root: infoStyles.dialogDelete }}>
+          Delete
+        </Button>
+        <Button size="small" classes={{ root: infoStyles.dialogButton }}>
+          Edit
+        </Button>
+      </DialogActions>
```
wongcmichelle: I think you need to style the buttons so they align with the left and right end of the div.

What it looks like now:
image

Figma:
image

wongcmichelle: Try `justify-content: space-between`
https://css-tricks.com/snippets/css/a-guide-to-flexbox/

src/components/Maps/InfoWindow.tsx
```diff
@@ -0,0 +1,82 @@
...
+      classes={{ root: infoStyles.root }}
+      open={open}
+      onClose={handleClose}
+      BackdropProps={{ style: { backgroundColor: 'transparent' } }}
+    >
+      <DialogContent style={{ overflow: 'hidden' }}>
+        <Typography variant="body1" color="textSecondary">
+          Name:
+          <span
+            style={{ display: 'inline', color: '#000000' }}
```
wongcmichelle: Do we have `#000000` in our `Colours` enum?

src/components/Maps/InfoWindow.tsx
```diff
@@ -0,0 +1,82 @@
...
+            color="textPrimary"
+          >
+            {` ${title} `}
+          </span>
+        </Typography>
+        <Typography variant="body1" color="textSecondary">
+          Location:
+          <span
+            style={{ display: 'inline', color: '#000000' }}
```
wongcmichelle: Do we have `#000000` in our `Colours` enum?

src/components/Maps/MapPage.tsx
```diff
@@ -0,0 +1,136 @@
...
+          map,
+          title: pin.label,
+        });
+
+        marker.addListener('click', () => {
+          map.setCenter(marker.getPosition());
+          setInfoWindowOpen(true);
+          setInterestPinLocation(pin.address);
+          setInterestPinTitle(pin.label);
+        });
```
wongcmichelle: In general, the `click` event works on desktop, but on an ipad it might not work. It doesn't work when I use the ipad pro via chrome inspection. You may have to try a touch event: https://reactjs.org/docs/events.html#touch-events.

src/graphql/queries/maps.ts
```diff
@@ -0,0 +1,30 @@
...
+  address: string;
+  createdAt: string;
+  updatedAt: string;
+}
+
+export const GET_PINS_BY_EVENT_ID = gql`
+  query pinsForEvent($eventId: ID!) {
+    pinsForEvent(eventId: $eventId) {
+      id
+      eventId {
```
wongcmichelle, line 19: Just want to confirm that this is okay given that event queries return more info.

wongcmichelle (changes requested):
Great work! Left some comments and I have a question. What are the pins for? From reading the notion ticket, my understanding was that the pins are for each CCP associated with the event. But i'm not seeing any CCPs when I got to the event dashboard for "Spooky Scary Event" even though I see pins when I go to the corresponding map page.

aaronabraham311:
> Great work! Left some comments and I have a question. What are the pins for? From reading the notion ticket, my understanding was that the pins are for each CCP associated with the event. But i'm not seeing any CCPs when I got to the event dashboard for "Spooky Scary Event" even though I see pins when I go to the corresponding map page.

You're right! The pins are the locations of CCPs and events but for the sake of convenience, I just made random pins that are not associated with any pre-existing CCPs or events. Sorry if I did not make that clear.

src/components/Maps/InfoWindow.tsx
aaronabraham311: Thank you! Will check that out.

src/components/Maps/InfoWindow.tsx
aaronabraham311: I have to double check. If it is, I'll just use the enum. If not, I'll add it in.

src/components/Maps/MapPage.tsx
aaronabraham311: This part of the code is not React code, this is code from Google Maps. For Google Maps, they recommend using the 'click' listener. I tested it out on the iPad and it works well.

src/graphql/queries/maps.ts
aaronabraham311: I believe this should be fine as we don't need any more information. All we really need is the ID for caching purposes. cc @petorsong what do you think?

petorsong: i'm ok with this as long as it doesn't have the potential to introduce caching issues in the future

this does bring up a good question on how we should structure our graphql queries (or which one is more of a graphql-ism)

we can either
1. have big/wide queries that fetch a lot of stuff from the backend (e.g. `events`, `patients`) and only do filtering for what we need in the frontend; this may force us to fetch a lot of data in cases where we don't necessarily need it
OR
2. have queries for very specific use cases (e.g. `ccpsForEvent`, `patientsForCcp`); this seems like it won't leverage apollo's caching capabilities to the fullest and may require more network requests in the long run

we're kind of in the middle of both options right now, lemme know if either of you have any opinions for either option

src/components/Maps/InfoWindow.tsx
petorsong: also don't forget the little X in the top right!

src/components/Maps/MapPage.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+const MapPage = ({
+  match: {
+    params: { eventId },
+  },
+}: {
+  match: { params: { eventId: string } };
+}) => {
+  const defaultMap = {
+    center: { lat: 43.470846, lng: -80.538473 },
```
petorsong: this will probably have to be tweaked and not hardcoded - potentially the coordinates of the `event` which the map can centre around?

src/components/Maps/MapPage.tsx
petorsong: tested out the current location and it was pretty off for me (maybe a km or 2 - dunno if it's just me though)

petorsong (changes requested):
this is really cool! also looks pretty good with the `iPad` screen size on firefox (but has a minor vertical scroll bar)
- @wongcmichelle made a really good point about pins and how they should be associated with collection points - the `locationPins` seeder should be removed and pin info should be included in the `collectionPoints` seeder instead
- i'm not sure i like the "selecting a pin centres the entire map around it" experience - i'll sync with rishabh to see if we want to change that and let you know

src/components/Maps/MapPage.tsx
aaronabraham311: Yep exactly. I will make a note of that in the upcoming event location ticket

src/components/Maps/MapPage.tsx
aaronabraham311: Mine is off as well but not by that large of a margin, more like a 100-200m. [This](https://developers.google.com/maps/documentation/javascript/examples/map-geolocation) is the recommended way of getting user's current location on Google Maps though. There is an [older version](https://developers.google.com/maps/documentation/geolocation/overview) of geolocation but because it's older, I wasn't sure if we should use it.

src/components/Maps/InfoWindow.tsx
```diff
@@ -47,33 +54,38 @@ const InfoWindow = ({
...
+                color="textPrimary"
+              >
+                {` ${address} `}
+              </span>
+            </Typography>
+        </DialogContent>
+        <CloseIcon onClick={handleClose} classes={{ root: infoStyles.closeButton }}/>
+      </Container>
+      <DialogActions style = {{ padding: '0px'}}>
+        <Container classes={{ root: infoStyles.flexContainer }}>
```
wongcmichelle: Can you add some padding for extra space below the buttons? (see Figma)

image

wongcmichelle (changes requested):
Thanks for addressing the comments :)

I noticed something I forgot to mention. I think the Maps icon should be black? It also looks different from Figma.

image

wongcmichelle:
@aaronabraham311 The map icon in Figma looks like this:

image

Also I meant padding below the button: 
image

In Figma it looks like:
image

src/components/Maps/MapPage.tsx
petorsong: my second try was right on the money, likely not a big issue

src/components/Maps/MapPage.tsx
```diff
@@ -0,0 +1,136 @@
...
+          };
+
+          const marker = new maps.Marker({
+            position: pos,
+            map,
+            title: 'Current position',
+          });
+
+          marker.setIcon(
+            'http://maps.google.com/mapfiles/ms/icons/green-dot.png'
```
petorsong: does this make a network request from google maps? is it possible to just use a `material-ui` asset/icon instead?

petorsong (changes requested):
ran into some usability issues (tested firefox dev tools on ipad setting)

- `InfoWindow` has a drop shadow? (is this intentional)
  - also right side padding is a bit too wide - figma mock looks like it should be the same all the way around
image
- `satellite` button seems to have some sort of dropdown menu below that's cut off
image
- satellite view and clicking on a pin makes it go back to normal map view
- `map`/`satellite` button flashes for a second when clicking a pin
- when `InfoWindow` is open, prevents any other button on the screen from being clicked (and repurposed instead to close `InfoWindow - kinda hijacks all other buttons on the screen: e.g. sidebar hamburger menu, map/satellite button, zoom buttons, etc.)

src/components/Maps/MapPage.tsx
aaronabraham311: We can make all our markers into `material-ui` icons but the issue is that we cannot link map functionality with the marker. For example, we currently make the map center on the pin that is clicked. If we were to use `material-ui` icons, we cannot use the `initMaps` function, thus blocking us from setting any sort of map functionality.

petorsong: actually forgot to mention before that the centering of the pin experience isn't one that me or rishabh like - it's very jarring to the user when selecting pins and can lead to users losing the context of where they were on the map before they selected a pin

is it possible to go back to the pattern of changing the colour of the selected pin instead of shifting the map to centre it?

aaronabraham311:
I made a lot of changes to this because of some things I needed for add pins. Here are the things that I changed:
- Marker design and colours: we are no longer panning to the centre of the map. Rather, markers that are not selected will be coloured lighter
- Styling changes for info window
- New `Marker` component
- Current location pin has changed

src/components/Maps/MapPage.tsx
```diff
@@ -0,0 +1,147 @@
+/* eslint-disable no-unused-vars */
```
petorsong: we shouldn't have this around, especially for a new component/file

src/components/Maps/MapPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+        open={infoWindowOpen}
+        handleClose={onInfoWindowClose}
+      />
+      <div style={{ height: '92vh', width: '100%', overflow: 'hidden' }}>
+        <GoogleMapReact
+          bootstrapURLKeys={{ key: process.env.REACT_APP_GMAPS }}
+          center={[defaultMap.center.lat, defaultMap.center.lng]}
+          zoom={defaultMap.zoom}
+          options={getMapOptions}
+          yesIWantToUseGoogleMapApiInternals
```
petorsong, line 121: flagging to remove?

src/components/Maps/Marker.tsx
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable */
```
petorsong: same deal here - this shouldn't be necessary, and if it's unavoidable it should be disabled on a line-by-line basis not for the entire file

petorsong (dismissed):
new changes are all great! have some nitpicks but functionally everything looks good, feel free to tackle these issues later in a cleanup ticket

- closing the info window is not very clean - it flashes without the pin-specific information for a sec before actually closing
Paramedics React - Mozilla Firefox 2020-11-02 21-08-20
- firefox usually polls me for a location, but it seems to be asking me for location access like every 5 seconds (this behaviour might be ok on an actual ipad, just making sure it's working as intended)
- the hamburger menu item would make more sense to be called `Map` instead of `Maps`

src/components/Maps/MapPage.tsx
aaronabraham311: I realized that we need to include this for us to access the state of the map type, i.e. is it hybrid version or roadmap version.

src/components/Maps/Marker.tsx
```diff
@@ -0,0 +1,47 @@
...
+  currentLocation: {
+    color: Colours.CurrentLocationMarker,
+  },
+  notClicked: {
+    color: Colours.MarkerNotClicked,
+  },
+});
+
+const Marker = ({
+  /* eslint-disable */
```
aaronabraham311, line 20: I need to include this to avoid linting errors. We need to pass in lat and lng because the API wrapper that we are using right now depends on us passing latitudes and longitudes to any child components of the map

src/styles/Constants.ts
```diff
@@ -26,4 +26,7 @@ export enum Colours {
   TriageBlack = '#000000',
   TriageSelectedBlack = '#BFBFBF',
   SnackbarButtonBlue = '#28B2FF',
+  Marker = '#EE502E',
+  MarkerNotClicked = '#F7B8AB',
```
wongcmichelle, line 30: nitpick
```suggestion
  MarkerNotSelected = '#F7B8AB',
```

aaronabraham311:
All other comments have been included in this [cleanup ticket](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3d4d369c668e4d2581a2517666768b44&p=b78b796088af404e9972165c0e366865) which I will target later.

wongcmichelle (approved):
Just one nitpick comment, otherwise looks good to me 🚢

src/styles/Constants.ts
aaronabraham311: I will include this in my cleanup ticket! Thanks for the catch.

src/components/Maps/Marker.tsx
petorsong: what's the linter error? seems kinda strange that we only want to exclude these 2 lines

petorsong (approved):
🚄

src/components/Maps/Marker.tsx
aaronabraham311: Linter complains that the variables are unused, but we need to pass it in

PR #47, opened 2020-10-28
Improve Dispatch Experience

patrick-du:
[Notion ticket](https://www.notion.so/uwblueprintexecs/Dispatch-experience-7a00c9866f244811a4993e4b79fd7583)

## Changes
- Implemented the functionality to update patient run number within the patient modal

## Testing
- Head to an event with a CCP that consists of a patient
- Click on a patient and update the `runNumber`
- Ensure that the button at the bottom right says `Edit` if there is no difference between the old run number and current input, otherwise, it should be `Save Changes`
- Click the `Save Changes` button

- Ensure that mutation occurred by heading to your GraphQL playground and querying for the patient you updated

before opening PR
- [X] check notion ticket
- [X] run linter
- [X] go through file diff

filling out PR
- [X] descriptive title
- [X] update `notion ticket` link
- [X] fill out `Changes`
- [X] fill out `Testing`
- [X] assign yourself to the PR

after opening PR
- [X] link PR to notion ticket
- [X] move ticket to `In PR`
- [X] make sure build passes
- [X] ping `@ps` in `#paramedics-dev`

### DIFF
--- a/src/components/CCPDashboard/PatientInfoTable.tsx
+++ b/src/components/CCPDashboard/PatientInfoTable.tsx
@@ -164,7 +164,6 @@ export const PatientInfoTable = ({
   const [runNumber, setRunNumber] = React.useState<number | null>(
     selectedPatient ? selectedPatient.runNumber : null
   );
-  const [isNewRunNumber, setIsNewRunNumber] = React.useState<boolean>(false);
   const history = useHistory();
   const { enqueueSnackbar } = useSnackbar();
 
@@ -233,8 +232,6 @@ export const PatientInfoTable = ({
 
   const handleRunNumber = (newRunNumber) => {
     const convertedRunNumber = newRunNumber ? parseInt(newRunNumber) : null;
-    const oldRunNumber = selectedPatient ? selectedPatient.runNumber : null;
-    setIsNewRunNumber(convertedRunNumber !== oldRunNumber);
     setRunNumber(convertedRunNumber);
   };
 
@@ -246,7 +243,6 @@ export const PatientInfoTable = ({
         collectionPointId: ccpId,
       },
     });
-    setIsNewRunNumber(false);
     enqueueSnackbar(`Patient ${selectedPatient?.barcodeValue} edited.`);
   };
 
@@ -416,7 +412,7 @@ export const PatientInfoTable = ({
               }`,
             }}
           >
-            {isNewRunNumber ? (
+            {runNumber !== selectedPatient?.runNumber ? (
               <Button
                 onClick={handleClickSave}
                 color="secondary"
--- a/src/components/CCPDashboard/PatientInfoTable.tsx
+++ b/src/components/CCPDashboard/PatientInfoTable.tsx
@@ -18,8 +18,14 @@ import {
 } from '@material-ui/core';
 import { useMutation } from '@apollo/react-hooks';
 import { MoreHoriz } from '@material-ui/icons';
+import { useSnackbar } from 'notistack';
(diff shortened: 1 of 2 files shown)


src/components/CCPDashboard/PatientInfoTable.tsx
```diff
@@ -155,9 +161,22 @@ export const PatientInfoTable = ({
...
+  const { enqueueSnackbar } = useSnackbar();
 
-  const [editPatient] = useMutation(EDIT_PATIENT);
+  const [editPatient] = useMutation(EDIT_PATIENT, {
+    update(cache) {
+      const patientId = ((selectedPatient as unknown) as Patient).id;
+      const { patients } = cache.readQuery<any>({
+        query: GET_ALL_PATIENTS,
+      });
+      setSelectedPatient(patients.find((x) => x.id === patientId));
```
patrick-du, line 176: This ensures that `selectedPatient` always contains a fresh state even when the modal is not closed and reopened. 

Without it, the `handleRunNumber` function will utilize the previous `runNumber` for comparison since `selectedPatient` is only set when the user opens a modal.

src/components/CCPDashboard/PatientInfoTable.tsx
```diff
@@ -155,9 +161,22 @@ export const PatientInfoTable = ({
     patients.find((x) => x.id === patientId)
   );
   const [anchorEl, setAnchorEl] = React.useState(null);
+  const [runNumber, setRunNumber] = React.useState<number | null>(
+    selectedPatient ? selectedPatient.runNumber : null
+  );
+  const [isNewRunNumber, setIsNewRunNumber] = React.useState<boolean>(false);
```
petorsong: is there a way to eliminate this variable? we already have `runNumber`, can't we just check that wherever we currently look at `isNewRunNumber` (i.e. just compare the saved run number with the one that's currently in the input field)?

src/components/CCPDashboard/PatientInfoTable.tsx
petorsong: just to double check, does this introduce any additional network requests by any chance? i checked my network tab and it seemed reasonable

petorsong (changes requested):
for next time, a screenshot or a short gif would be helpful for reviewers to understand what to expect

added a couple of comments but otherwise looks fine to me!

src/components/CCPDashboard/PatientInfoTable.tsx
patrick-du: Nope - I don't believe so, according to this https://www.apollographql.com/docs/react/caching/cache-interaction/#readquery
image

src/components/CCPDashboard/PatientInfoTable.tsx
patrick-du: Yeah, that makes sense. Thanks, I've made the change in the latest commit.

petorsong (approved):
🏎️

jesskeepswimming:
image
image
image

All non-edge cases look good-- however, I noticed that changing the run number to an edge case such as 76-54 will show that changes are saved when the run number actually get clears. I'm not sure if this edge case will realistically get encountered though!

jesskeepswimming (approved):
image
image
image

All non-edge cases look good-- however, I noticed that changing the run number to an edge case such as 76-54 will show that changes are saved when the run number actually get clears. I'm not sure if this edge case will realistically get encountered though!

petorsong:
@jesskeepswimming run numbers will always be numeric - no need to test inputs that contain other characters

PR #49, opened 2020-11-04
Fixed inconsistent dates for different event pages

sarafong:
[Notion ticket](https://www.notion.so/uwblueprintexecs/Edit-Event-Page-Event-calendar-display-s-today-s-date-not-the-event-date-86c88f5c2a40478a830f66628bb1fedb)

## Changes
- Because we only collect month/day, time zone gets messed up when parsing from the db. `EventCreationPage` should now retrieve the date as a string from the DB (instead of parsing it into a Date object as before). 
- Date picker will select as a date but pass to DB as string in the format `yyyy-mm-dd`. 
- Fixed off by one error for month label array

## Screenshots

These three views should be the same: 

Screen Shot 2020-11-03 at 11 57 56 PM
Screen Shot 2020-11-03 at 11 58 17 PM
Screen Shot 2020-11-03 at 11 58 28 PM

## Testing
- Verify event dates are the same on home page, event page, and edit event page
- Try adding or editing an event and making sure the date you select is the date that's set. Also check in backend to make sure it looks correct in the db.

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [ ] update `figma link`
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [x] make sure build passes


sarafong:
After I made the PR I found a bug with my fix when you create a new event so please hold off on reviewing :(

sarafong:
ok it's good to go now! @petorsong @patrick-du

petorsong (changes requested):
this date business is kind of confusing me and i'm not sure how we want to approach this from a UX perspective - let's pull rishabh in and chat next work session

* i tested this at about 8 pm PST and i got a mismatch between my selected date and the displayed date
image
* also the `moment` library will no longer be maintained as of sep 2020, so i would prefer to not introduce it into any new code - this has some suggestions for how to approach that
  * https://momentjs.com/docs/#/-project-status/

sarafong:
yikes ok looks like it's still mismatched, need to test this further. also very unfortunate that moment is being discontinued :(

sarafong:
ready to be reviewed again!

src/components/EventCreation/EventCreationPage.tsx
```diff
@@ -115,13 +115,11 @@ const EventCreationPage = ({
...
-        (obj, currentPart) => ({
-          ...obj,
-          [currentPart.type]: currentPart.value,
-        }),
-        {}
-      )
+    ? {
+        year: eventDate.split('-')[0],
+        month: eventDate.split('-')[1],
+        day: eventDate.split('-')[2],
```
petorsong, line 121: do we need to add the `literal` field as well?

petorsong (approved):
seems to be working now whoo

patrick-du:
Nice work! 

Not sure why but I'm getting a mismatch in the `Edit Event` page between the date that's in the input field and the date displayed in the date picker. 

Besides this, the rest of the PR looks good and the dates are consistent between the other screens.
ezgif com-gif-maker (1)

patrick-du (approved):

PR #55, opened 2020-11-13
Add Location Pin

aaronabraham311:
[Notion ticket](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3d4d369c668e4d2581a2517666768b44&p=03d6681d60694dbeb4612d088595b3f9)

[Figma link](https://www.figma.com/file/IpOPyv79gBwEohvOJvPV6f/%F0%9F%9A%91-Designs-for-Development-S20-F20?node-id=1174%3A21019)

## Changes
- Add pin flow added. This includes:
    - Add pin button
    - Sidebar with autocompletion of addresses (restricted to Canada)
- Enhanced geolocation
- Adding a few more queries
- Minor changes to the hamburger menu due to styling issues

## Screenshots
Screen Shot 2020-11-13 at 1 33 38 PM

## Testing
- Test the happy path: make sure that pins are added and open the resulting info window to make sure the information is correct
- Try partially filling in fields and hit cancel. Reopen and all fields should be cleared
- Partially fill fields and tap outside (not hitting cancel). Reopen and fields (except address) should be there
- Try to enter an address outside of Canada. This should not work
- Testing on iPad via tunnelling will not work on this PR because tunnelling wipes the API key
- Check the cache and confirm that it is updating
- Make sure this all works with the hybrid setting 

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] update `figma link`
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [ ] make sure build passes


src/components/Maps/MapPage.tsx
```diff
@@ -76,15 +82,34 @@ const MapPage = ({
...
-      });
+      navigator.geolocation.getCurrentPosition(
+        (position: Position) => {
+          setCurrentLocationPin({
+            lat: position.coords.latitude,
+            lng: position.coords.longitude,
+          });
+        },
+        geolocationErrorCallback,
+        { enableHighAccuracy: true }
```
aaronabraham311, line 93: We are trying this option for enhanced geolocation. Let me know if this causes issues.

src/components/Maps/SearchBar.tsx
```diff
@@ -0,0 +1,130 @@
...
+  onComplete,
+}: {
+  onComplete: ({ latitude, longitude, address }) => void;
+}) => {
+  const [address, setAddress] = useState('');
+  const [menuOpen, setMenuOpen] = useState(false);
+  const styles = useStyles();
+  const inputEl = useRef(null);
+
+  const searchOptions = {
```
aaronabraham311, line 43: This restricts our search to Canada

src/components/common/MenuAppBar.tsx
```diff
@@ -119,7 +119,12 @@ export default function MenuAppBar(props: MenuAppBarProps) {
           </Typography>
         </Toolbar>
       </AppBar>
-      <Drawer anchor="left" open={isMenuOpen} onClose={toggleMenu}>
+      <Drawer
+        anchor="left"
+        open={isMenuOpen}
+        onClose={toggleMenu}
+        disableScrollLock
```
aaronabraham311, line 126: This was shifting a lot of buttons as it was adding right padding to our `body` container.

src/components/Maps/MapPage.tsx
```diff
@@ -31,13 +35,15 @@ const MapPage = ({
...
   const [mapTypeId, setMapTypeId] = React.useState(MapTypes.ROADMAP);
+  const [openSidebar, setOpenSidebar] = React.useState(false);
   const defaultMap = {
     zoom: 11,
     center: {
       lat: 43.470846,
       lng: -80.538473,
     },
   };
+  const geolocationErrorCallback = undefined;
```
petorsong, line 46: in general `null` is a better value to use than `undefined` - `undefined` should be reserved for something that is actually undefined

with that in mind, why are we not setting this variable in the first place?

src/components/Maps/MapPage.tsx
petorsong: do you know if there are any differences that we should expect with and without this flag?

src/components/Maps/Sidebar.tsx
```diff
@@ -0,0 +1,127 @@
...
+            longitude,
+            address,
+          });
+          setAddress('');
+          setLabel('');
+          e.preventDefault();
+        }}
+      >
+        <Typography variant="body1" classes={{ root: styles.label }}>
+          {' '}
```
petorsong: what is this doing? same on line 95

src/components/Maps/Sidebar.tsx
```diff
@@ -0,0 +1,127 @@
...
+          errorMessages={['A pin name is required']}
+          className={styles.pinLabelField}
+        />
+        <Typography variant="body1" classes={{ root: styles.label }}>
+          {' '}
+          Pin Location:
+        </Typography>
+        <SearchBar
+          onComplete={({ latitude, longitude, address }) =>
+            handleGeocodeSelection({
```
petorsong, line 98: can you explain at a high level how this geocode and autocomplete libraries/APIs are functioning in our code?

src/components/Maps/AddPinButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const classes = useStyles();
+  return (
+    <div>
+      <Button
+        className={classes.addButton}
+        variant="contained"
+        color="secondary"
+        onClick={handleClick}
+      >
+        <Add className={classes.buttonIcon} />
```
JasonYG, line 29: Not your fault, but this component name is so wack lol, mui should've named this `AddIcon` or something

src/components/Maps/InfoWindow.tsx
```diff
@@ -64,6 +64,7 @@ const InfoWindow = ({
       }}
       hideBackdrop
       disableEnforceFocus
+      disableScrollLock
```
JasonYG, line 69: I'm just guessing that this allows the user to scroll up and down in the InfoWindow, right?

src/components/Maps/MapPage.tsx
JasonYG, line 46: we're going to define this when we tackle error handling right? If so, maybe add a TODO here so we don't forget

src/components/Maps/SearchBar.tsx
```diff
@@ -0,0 +1,130 @@
...
+  };
+
+  return (
+    <PlacesAutocomplete
+      value={address}
+      onChange={handleChange}
+      onSelect={handleSelect}
+      searchOptions={searchOptions}
+    >
+      {({ getInputProps, suggestions, getSuggestionItemProps, loading }) => (
```
JasonYG, line 82: i might've missed it, but where do we define `suggestions` and `getSuggestionItemProps`? Or is this provided by the library we're using?

JasonYG (commented):
Reviewed the code, looks good! Left a few comments, but nothing big.

Would you be down to test this PR with me during our Tuesday work session, so you can walk me through some of the changes?

src/components/Maps/Sidebar.tsx
```diff
@@ -0,0 +1,127 @@
...
+    setLongitude(lng);
+    setAddress(address);
+  };
+
+  return (
+    <Drawer
+      open={open}
+      onClose={onClose}
+      PaperProps={{ style: { width: '400px' } }}
+      disableScrollLock
```
petorsong, line 64: what exactly does this do?

petorsong (changes requested):
nice, glad to see we got pin adding working! i went through the changes with stacy and i'll add more styling/design related ones to your cleanup ticket for now

- info window is aligned differently and has different sizes - differences between the ones loaded by the seeder and the ones i added via this PR (may be an issue with the string sizes within the component?)
- console errors when `react--places-autocomplete` can't fetch any results - shouldn't surface the error there
- is it possible to add an empty state for when there are no suggestions?
- coordinates not entirely accurate - i added `waterloo city hall` as a pin and took the coordinates from the db on actual google maps to compare, and the pins were slightly different; do we know why this is the case?

src/components/Maps/MapPage.tsx
aaronabraham311: I don't notice much of difference. Latency was also not affected.

src/components/Maps/InfoWindow.tsx
aaronabraham311: `disableScrollLock` allows the user to scroll the background when modals are open. Weirdly enough, if there is a scroll lock, then padding is added to the body

src/components/Maps/MapPage.tsx
aaronabraham311: Yeah I wasn't sure what our error handling policy was. I will discuss with Peter about this.

src/components/Maps/SearchBar.tsx
aaronabraham311: This is from the `react-places-autocomplete` library

src/components/Maps/Sidebar.tsx
aaronabraham311: Weird, prettify must have added that. I will remove that.

src/components/Maps/Sidebar.tsx
aaronabraham311: `react-places-autocomplete` provides the following functions:
- `getInputProps`: these are the props that are passed to input elements. We need to pass this in so that the component can track our inputs and our selections
- `suggestions`: these are the suggestion objects returned from [Google Maps](https://developers.google.com/places/web-service/autocomplete#place_autocomplete_responses)
- `getSuggestionItemProps`: similar to `getInputProps`, this just helps the component keep track of our components that actually display our suggestions
- `geocodeByAddress`: we get the latitude and longitude of a particular address.

Whenever we type in anything in the input textfield, the library puts a request for the autocompletion suggestions. These suggestions are then displayed. Upon selection, we geocode and pass our coordinates to other components.

I will be writing up documentation when I have finished the most crucial parts of this feature.

src/components/Maps/Sidebar.tsx
aaronabraham311: Jason asked this: https://github.com/uwblueprint/paramedics-react/pull/55#discussion_r523817420

aaronabraham311:
> coordinates not entirely accurate - i added waterloo city hall as a pin and took the coordinates from the db on actual google maps to compare, and the pins were slightly different; do we know why this is the case?

I checked it and it is off by a few decimals. I think this is from the API call rather than from any library. Also, the difference is quite small so I don't imagine this being a huge problem

Latitude and longitude of city hall from API call:
Screen Shot 2020-11-15 at 3 10 08 PM

Latitude and longitude of city hall from Google:
Screen Shot 2020-11-15 at 3 11 30 PM

aaronabraham311:
> is it possible to add an empty state for when there are no suggestions?

I initially had this implementation but I think it will look ugly. I'll talk with Stacy about this but I am leaning against it.

JasonYG (approved):
Changes look good to me!

petorsong (approved):
🔍

PR #56, opened 2020-11-18
Login and Unauth Landing Screens

JasonYG:
[Notion ticket](https://www.notion.so/uwblueprintexecs/9224caeb21ce4350b8a1dc378d393323?v=43f44ac04d734394b7f7edc636c523b4)

[Figma link](https://www.figma.com/file/IpOPyv79gBwEohvOJvPV6f/%F0%9F%9A%91-Designs-for-Development-(S20%2C-F20)?node-id=1972%3A36028)

## Changes
- Added unauth/login page
- Added static images (paramedics logo and ambulance photo) to page, both of which are subject to change
- Added new routes (`/login` and `/unauth`) that redirect to the page

## Screenshots
`/login`
image

`/unauth` - The user would be redirected here if they logged in and their email wasn't in our Users table
image

## Testing
- Verify that the page looks right on iPad sizing

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] update `figma link`
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [x] make sure build passes


src/components/EventOverview/UserProfile.tsx
```diff
@@ -35,6 +35,7 @@ const UserProfile = () => {
     // TODO: Redirect to edit profile
   };
   const handleLogoutClick = (event) => {
+    fetch("http://localhost:4000/logout");
```
JasonYG: this will eventually logout the current user, but i'm running into a CORS issue right now

JasonYG: I removed this line in the latest commit to prevent errors - this will eventually be added back in when I tackle logout functionality

src/components/Maps/SearchBar.tsx
```diff
@@ -116,16 +116,17 @@ const SearchBar = ({
...
+                suggestions.map((suggestion) => (
+                  <ListItem
+                    key={suggestion.description}
+                    classes={{ root: styles.sidebarOptions }}
+                  >
+                    <div {...getSuggestionItemProps(suggestion)}>
+                      <Typography>{suggestion.description}</Typography>
+                    </div>
+                  </ListItem>
+                ))}
```
petorsong: let's revert this, i'd rather not have random formatting changes check in especially if the other changes within the commit are unrelated

src/components/Login/LoginButton.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import { makeStyles, Button } from '@material-ui/core';
+
+const useStyles = makeStyles({
+  addButton: {
+    padding: '12px 23px',
+  },
+});
+
+const LoginButton = () => {
```
petorsong, line 11: is it possible to use our common button for this? seems like it's not doing anything fancy

src/components/Login/InvalidUserMessage.tsx
```diff
@@ -0,0 +1,50 @@
...
+    height: '19px',
+  },
+  invalidUserText: {
+    display: 'inline-block',
+    paddingLeft: '19.7px',
+    paddingBottom: '7px',
+  },
+});
+
+const InvalidUserMessage = () => {
```
petorsong, line 27: we can fold this component into `LoginPage` since it's the only place it'll be used anyways

src/components/Login/LoginPage.tsx
```diff
@@ -0,0 +1,95 @@
...
+    <Box className={classes.root}>
+      <Box width="50%" className={classes.leftColumn}>
+        <Box paddingLeft="122px">
+          <Typography variant="h3">Welcome</Typography>
+          <Typography variant="body1" className={classes.boldText}>
+            First time here?
+          </Typography>
+          <Typography variant="body1" className={classes.adminAccessText}>
+            Check that your administrator has granted you access before logging
+            in
```
petorsong, line 68: ?
```suggestion
            Check that your administrator has granted you access before logging in
```

petorsong (changes requested):
naisu - looks like we're just scaffolding this page out and it doesn't do anything quite yet; comments are mostly not complicated
- we shouldn't check in a package-lock change unless there's a corresponding package.json change - please remove the changes from this PR
- images should go into an `src/assets/` directory
  - also we literally have the transparent image? wouldn't it be better to have the original image and manage transparency ourselves? lol

src/components/Login/InvalidUserMessage.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <Typography
+          color="error"
+          className={classes.invalidUserText}
+          variant="caption"
+        >
+          Invalid User
+        </Typography>
+      </Box>
+      <Typography color="error" variant="caption">
+        Sorry, you don’t have access to the system yet! Please see administrator
```
aaronabraham311, line 42: I am assuming the line break is from the linter?

aaronabraham311:
Looks great! The sizing on the iPad was messed up when I checked the `unauth` screen.
Screen Shot 2020-11-23 at 3 39 55 PM

src/components/Login/LoginPage.tsx
aaronabraham311: I think this is the linter. Seems like it has no functional impact

src/components/Login/InvalidUserMessage.tsx
petorsong: neither of these should hit the [80 char limit which is what prettier enforces](https://github.com/uwblueprint/paramedics-react/blob/master/.prettierrc#L2) - i think it hurts readability so let's just make sure static strings can stay on one line (when short enough like these)

src/components/Maps/SearchBar.tsx
JasonYG: hmm i think the linter (`npm run lint`) did this though, and if I revert I think it won't pass the `dev-lint` test

JasonYG: somebody might've forgotten to lint their files before merging, which is why this change is being made. I think a similar thing happened with Ricky's recent PR. That being said, should I still revert?

src/components/Login/LoginButton.tsx
JasonYG: not sure if we have a common button that handles this case - I only see a `BackButton` and a `CancelButton` in our common components folder.

src/components/Login/InvalidUserMessage.tsx
JasonYG: hmm yeah, but isn't that true for all components defined within a specific page's folder? e.g. `AddBinButton` will only be used for the Maps page, `StatusPills` will only be used for the PatientProfile page, etc. I think keeping this separate might be cleaner, since there's a bunch of custom styling for this component. 

Thoughts?

src/components/Login/LoginPage.tsx
JasonYG: yup it's the linter, I tried making this one line, but running `npm run lint` forces it to be two

src/components/Login/InvalidUserMessage.tsx
JasonYG: yeah this change is being caused by the linter. I tried moving it to one line, but `npm run lint` forces it to be two. Maybe the linter (eslint or prettier) counts the white space indents in the char count?

JasonYG:
> Looks great! The sizing on the iPad was messed up when I checked the `unauth` screen.
> Screen Shot 2020-11-23 at 3 39 55 PM

yeah good catch, I think i noticed this too when I was testing the changes. The problem is that i copied the sizing from figma (like the width and height in px), which was probably designed with desktop resolution.

Edit: I just talked with Stacy, and she said she'd redo some of the sizing on Figma.

JasonYG:
> * also we literally have the transparent image? wouldn't it be better to have the original image and manage transparency ourselves? lol

good point about the transparency image, I just downloaded it off Figma like Stacy told me and it gave me the transparent image lol

Regarding the `src/assets` folder, I agree i think we should place all images in this folder. The reason I didn't was because we already had images in the root directory (`map.png`, `logo.svg`), so I thought that was the convention we were following. Should I move all these static elements to a `assets` folder?

petorsong:
> 
> 
> > * also we literally have the transparent image? wouldn't it be better to have the original image and manage transparency ourselves? lol
> 
> good point about the transparency image, I just downloaded it off Figma like Stacy told me and it gave me the transparent image lol
> 
> Regarding the `src/assets` folder, I agree i think we should place all images in this folder. The reason I didn't was because we already had images in the root directory (`map.png`, `logo.svg`), so I thought that was the convention we were following. Should I move all these static elements to a `assets` folder?

yes please (note that `map.png` shouldn't be there much longer since it is a placeholder for our gmaps stuff)

aaronabraham311 (dismissed):
Nice work!

src/components/Maps/SearchBar.tsx
aaronabraham311: I made sure I linted SearchBar before merging it in...this should not be happening.

petorsong (approved):

PR #64, opened 2020-11-25
Edit and delete location pins

aaronabraham311:
[Notion ticket](https://www.notion.so/uwblueprintexecs/3-Edit-and-delete-location-pins-233db29d406e4fc79aed2b6b660d80f6)

[Edit pin Figma link](https://www.figma.com/file/IpOPyv79gBwEohvOJvPV6f/%F0%9F%9A%91-Designs-for-Development-S20-F20?node-id=1174%3A21142)
[Delete pin Figma link](https://www.figma.com/file/IpOPyv79gBwEohvOJvPV6f/%F0%9F%9A%91-Designs-for-Development-S20-F20?node-id=1174%3A21284)

## Changes
- Added edit pin flow
    - Click on a pin -> click edit -> change it up and confirm -> pin has been edited!
    - Info window doesn't close if you click edit (I thought this would be useful for the paramedics since they can continue to refer to the pin even though it has been edited
- Added delete pin flow
    - Canceling pin deletion doesn't close info window (again, same reason as above)
- Changed a bit of confirm modal as it was calling `handleCancel`, which was messing up the expected behaviour of canceling a deletion
## Screenshots
About to edit a pin
Screen Shot 2020-11-24 at 6 10 04 PM
After edit completion:
Screen Shot 2020-11-24 at 6 08 13 PM

Deletion (not completely updated in screenshot, but information window will be open at this time according to Figma:
Screen Shot 2020-11-24 at 6 09 37 PM

## Testing
- Test the happy paths:
    - Edit a pin and make sure the changes are properly reflected (change both the name of the pin as well as the location)
    - Delete a pin and make sure the changes are properly reflected on both the front end and the backend
- Make sure cancellation functionality is working: for both edit and delete, canceling the flow should leave the info window open
- Create three pins and perform the following:
    - 3 consecutive deletes
    - 3 consecutive edits
    - Edit a pin -> click add (fields should be cleared) -> completion/cancellation should clear the info window (you should not see it)
    - Delete -> edit -> add
    - Any other combinations

**Note: If you have any suggestions/improvements on my state management, I will gladly take it!! It seems messy to me**

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] update `figma link`
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [x] make sure build passes


src/components/Maps/MapPage.tsx
```diff
@@ -95,6 +100,17 @@ const MapPage = ({
     }
   });
 
+  useEffect(() => {
```
aaronabraham311, line 103: I was running into issues with delete modal cancellation. Essentially, once I confirmed deletion, the info window was  still open. After lots of testing with @JasonYG, we decided to go with a `useEffect` solution and remove the usage of `handleCancel` for the maps confirmation modal.

src/components/Maps/MapPage.tsx
```diff
@@ -199,7 +302,22 @@ const MapPage = ({
...
+      />
+      <ConfirmModal
+        open={isDeleteClicked}
+        handleClickCancel={onDeletePinCancel}
+        handleClickAction={onDeletePinConfirm}
+        title="You are about to delete a location pin."
+        body="Deleted location pins will no longer be accessible to other team members. Are you sure you want to delete this pin?"
+        actionLabel="Delete"
+        isActionDelete
+        isMap
```
aaronabraham311: Related to above comment

src/components/common/ConfirmModal.tsx
```diff
@@ -96,7 +98,9 @@ const ConfirmModal = ({
           onClick={(event) => {
             event.stopPropagation();
             handleClickAction();
-            handleClickCancel();
```
aaronabraham311: Related to above comment regarding deletion modal issues

src/components/common/ConfirmModal.tsx
```diff
@@ -58,12 +58,14 @@ const ConfirmModal = ({
   title,
   body,
   actionLabel,
+  isMap,
```
petorsong: i don't like introducing new props in such a widely used common component, especially if it's so specific to a particular use case - is there no other way around this? e.g. pass a no-op into `handleClickAction={() => {}}` or something of the like - i'd rather not build in one-off changes to address hyper-specific use cases in common components in general

src/components/Maps/SearchBar.tsx
```diff
@@ -31,11 +31,13 @@ const useStyles = makeStyles({
...
 
 const SearchBar = ({
+  editAddress,
   onComplete,
 }: {
+  editAddress?: string;
   onComplete: ({ latitude, longitude, address }) => void;
 }) => {
-  const [address, setAddress] = useState('');
+  const [address, setAddress] = useState('' || editAddress);
```
petorsong: nit: default states should go last
```suggestion
  const [address, setAddress] = useState(editAddress || '');
```

src/components/Maps/MapPage.tsx
petorsong: this isn't great but if it's the only option we can leave it in; still want to avoid messing with common components (see `ConfirmModal` comment)

petorsong (changes requested):
looks good overall! had some gripes with the `ConfirmModal` changes, would be great if we can find another approach
- i was able to create a pin that doesn't render on the frontend - it created a row in `locationPins` with every field EXCEPT address (so it had coordinates and an `eventId`) but it doesn't show up on the map
  - i couldn't repro at all but might be worth attempting to repro on your end

src/components/common/ConfirmModal.tsx
aaronabraham311: As discussed, this will be resolved by [this ticket](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=d7885c94b20c426cb1258dedac09dd12)

aaronabraham311:
Rolled back the deletion changes I have made to confirm modal. This means I will be anticipating that Patrick will fix the confusion with `ConfirmModal` opening on deletion. This also means that the modals for deletion is a little bit messed up as the info window is still open after deletion.

Also made some changes so that editing and deletion is compatible with [this backend branch](https://github.com/uwblueprint/paramedics-web/pull/83)

petorsong (approved):

PR #70, opened 2020-12-08
Map tap functionality

aaronabraham311:
[Notion ticket](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3d4d369c668e4d2581a2517666768b44&p=35bf1c9c71c74ca397d6240546e79b31)

## Changes
- Once a user taps on a particular location on the map, the 'Add Pin' sidebar will open with the corresponding address
- If a user adds a pin using the 'Add Pin' sidebar, then they have the option to tap on the map to designate the location for the pin. If they select a suggestion, a pin should appear at the location
- Tapping is now integrated to edit a pin's location (i.e. if you want to edit a pin's location, you can go through the same flow as listed above and simply tap the map to change the location)
- Marker should not shift anymore as you move around the map
- Integrated geocoding to convert latitude and longitude to addresses

## Screenshots
Kapture 2020-12-08 at 13 08 16

## Testing
**Please use ngrok to test on an iPad!!**
- Add a pin via button
- Add a pin via tap
- Add a pin via button and change address. The suggestion you click should be what is filled in the 'Pin Location' field
- Do above but use tap
- Edit a pin and change the location using tap
- Edit a pin and change the location using the autocompletion field
- Move around map and make sure markers do not move (esp. zooming in and out)

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] update `figma link`
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [x] make sure build passes


src/components/Maps/Marker.tsx
```diff
@@ -11,8 +11,17 @@ import eventPin from '../../assets/eventPin.svg';
 import otherPin from '../../assets/otherPin.svg';
 
 const useStyles = makeStyles({
+  root: {
+    textAlign: 'center',
+    position: 'absolute',
+    transform: 'translate(-50%, -50%)',
```
aaronabraham311, line 17: These styling changes should prevent the marker from shifting as you zoom in and out

src/components/Maps/Marker.tsx
```diff
@@ -44,7 +53,13 @@ const Marker = ({
     return <RadioButtonCheckedIcon className={styles.currentLocation} />;
   }
   return (
-    <Icon onClick={onClick} fontSize="large">
+    <Icon
+      onClick={onClick}
+      fontSize="large"
+      className={styles.root}
+      onMouseDown={(e) => e.stopPropagation()}
```
aaronabraham311, line 60: This should prevent any pin clicks from bubbling down as map clicks. Please let me know if this doesn't work for you!

petorsong (commented):
- tapping - pin colour is not the dark red indicating the currently selected pin
- editing a pin - the `add pin` button still shows up on the map (and blocks a good chunk of the screen)
- nit: tapped pin doesn't centre around the visible portion of the screen - assumes the map is the full screen size so is heavily left aligned in the map viewport (i believe this is already covered in a cleanup task)
- nit: we don't centre back to the original pin location if we attempt to edit a pin by clicking elsewhere and exit the edit side bar (this can be thrown into cleanup - more quality of life)

src/components/Maps/Sidebar.tsx
```diff
@@ -62,14 +72,25 @@ const Sidebar = ({
...
     }
 
-    if (editAddress) {
+    if (editAddress && editLocation && address === '') {
       setAddress(editAddress);
+      setLatitude(editLocation.lat);
+      setLongitude(editLocation.lng);
     }
-  }, [editLabel, editAddress]);
+  }, [clickedAddress, clickedLocation, editLabel, editAddress]);
```
petorsong: image
obligatory please double check this works instead of making any assumptions
```suggestion
  }, [clickedAddress, clickedLocation, editLabel, editAddress, editLocation, address]);
```

petorsong (dismissed):
one warning but good otherwise

petorsong (approved):

carelynntsai (approved):
so cool!

PR #72, opened 2020-12-13
Added wrapped text to patient's profile page

rickymao:
[Notion ticket](https://www.notion.so/uwblueprintexecs/wrap-the-text-for-the-notes-field-065184f2040043df86b4a3022abf48c0)

[Figma link](https://www.notion.so/uwblueprintexecs/wrap-the-text-for-the-notes-field-065184f2040043df86b4a3022abf48c0#f6cb2982df004e1582409d31fdba6dce)

## Changes
- Added text wrapping to patient's page

## Screenshots
Screen Shot 2020-12-12 at 4 53 10 PM

## Testing
- Opened up a patient's page in edit patient and typed a really long paragraph.
- Checked if other fields are affected by the new change

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] update `figma link`
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [x] make sure build passes

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -46,11 +46,7 @@ function App() {
         path="/events/:eventId/edit"
         component={(props) => <EventCreationPage mode="edit" {...props} />}
       />
-      <Route
-        exact
-        path="/manage/:resource"
-        component={ResourceOverviewPage}
-      />
+      <Route exact path="/manage/:resource" component={ResourceOverviewPage} />
       <Route exact path="/manage/">
         <Redirect to="/manage/members" />
       </Route>
--- a/src/components/common/FormField.tsx
+++ b/src/components/common/FormField.tsx
@@ -12,10 +12,13 @@ const useTextFieldStyles = makeStyles({
     padding: '20px',
     marginTop: '0px',
     marginBottom: '20px',
-    height: '120px',
+    minHeight: '120px',
     width: '100%',
     '& .MuiInput-formControl': {
       marginTop: 'auto',
+    },
+    '& .MuiInputBase-multiline': {
+      marginTop: '43.625px',
     },
     '& label': {
       fontWeight: 'bold',
@@ -47,6 +50,7 @@ const FormField: React.FC<{
   readOnly?: boolean;
   disabled?: boolean;
   numeric?: boolean;
+  isMultiline?: boolean;
 }> = ({
   label,
   placeholder,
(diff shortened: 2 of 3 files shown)


src/components/common/FormField.tsx
```diff
@@ -47,6 +50,7 @@ const FormField: React.FC<{
   readOnly?: boolean;
   disabled?: boolean;
   numeric?: boolean;
+  isMultiline?: boolean;
```
petorsong, line 53: nit: don't have to change it here, but at this point `FormField` is getting pretty clunky with a lot of props that are simply passed through to the material UI `TextField` component - instead of catching them explicitly and passing it through here, we should just take in additional props to `FormField` and spread them into `TextField`; this lets us catch only the props we care about in this component (which there are very few) and handle them accordingly while also allowing us to pass relevant props to `TextField` without worrying about them

again not necessary for this PR but if we need to add another prop to this component, we should consider this approach

petorsong (approved):
🏞️

src/App.tsx
```diff
@@ -46,11 +46,7 @@ function App() {
         path="/events/:eventId/edit"
         component={(props) => <EventCreationPage mode="edit" {...props} />}
       />
-      <Route
-        exact
-        path="/manage/:resource"
-        component={ResourceOverviewPage}
-      />
+      <Route exact path="/manage/:resource" component={ResourceOverviewPage} />
```
aaronabraham311, line 49: This is probably a prettify change. Are we still good with it in this PR?

aaronabraham311 (commented):
Not sure whether this is in the scope of this PR but there is no multiline in the patient modal on the CCP dashboard. If it's easy-enough change, we could just do it in this PR. What do you think? @rickrm @petorsong 
Screen Shot 2020-12-13 at 12 46 52 PM

src/App.tsx
petorsong: yes - this means that whoever checked this in forgot to run prettier before merging and if we don't address this here then it'll keep showing up on other people's PRs (after they run prettier)

petorsong:
> Not sure whether this is in the scope of this PR but there is no multiline in the patient modal on the CCP dashboard. If it's easy-enough change, we could just do it in this PR. What do you think? @rickrm @petorsong

there's no need to have a text area in the patient modal as multiline is specifically meant for input fields, not static text; that being said, it seems you have a very long word (with no spaces) but it works for a long string with a bunch of words - i'm fine with just supporting that for now since my guess is that having long words is far less likely than having a long string
image

aaronabraham311 (approved):

PR #76, opened 2020-12-31
Event/CCP flow via maps

aaronabraham311:
[Notion ticket](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3d4d369c668e4d2581a2517666768b44&p=82c893cef178429583046605fd4e9535)

[Figma link](https://www.figma.com/file/IpOPyv79gBwEohvOJvPV6f/%F0%9F%9A%91-Designs-for-Development-S20-F20?node-id=3767%3A0)

## Changes
- Adding and editing CCPs and events will now use the Map component. This ticket was mostly migrating queries and mutations into the `MapPage` component and rendering fields accordingly
- Removed `EventCreation` and `CCPForm` components as all functions would have been covered by `MapPage`
- Disabled editing and deleting `CCP` and `EVENT` pins on the map (must go through flow)
- Removed address placeholders throughout app and used pin address instead
- Added a much cleaner date picker for event flows
- Cascade deletion of location pins when deleting CCPs

## Screenshots
Adding events (with date picker open)
Screen Shot 2020-12-31 at 4 25 43 PM

Editing events
Screen Shot 2020-12-31 at 4 26 43 PM

Adding CCP
Screen Shot 2020-12-31 at 4 27 34 PM

Editing CCP
Screen Shot 2020-12-31 at 4 27 59 PM

Event dashboard now has the event address
Screen Shot 2020-12-31 at 4 28 18 PM

So does the CCP dashboard
Screen Shot 2020-12-31 at 4 28 42 PM

## Testing
- Add an event -> check the event map and ensure that the event pin is there
- Edit the event -> make sure the changes persist in the event description in the event dashboard and map
- Delete the event
- Do everything above but for CCPs
- Go through the same flow for regular pins (i.e. add a pin on the map itself) and ensure you can edit and delete the pin on the map itself

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] update `figma link`
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [x] make sure build passes


src/components/EventDashboard/CCPTabPanel.tsx
```diff
@@ -167,6 +168,25 @@ const CCPTabPanel = ({ eventId }: { eventId: string }) => {
         variables: { eventId },
         data: { collectionPointsByEvent: updatedCCPsList },
       });
+
+      // Update GET_PINS_BY_EVENT_ID
```
aaronabraham311, line 188: Cascade deletion of associated pins when deleting a CCP

src/components/Maps/InfoWindow.tsx
```diff
@@ -104,14 +107,24 @@ const InfoWindow = ({
         <Container classes={{ root: infoStyles.flexContainerButtons }}>
```
aaronabraham311, line 107: All changes in this file is to prevent the user from editing a `CCP` or `EVENT` pin while they are on the map

src/components/Maps/Sidebar.tsx
```diff
@@ -37,30 +63,67 @@ const useStyles = makeStyles({
...
+      dateParts.day = '0'.concat(dateParts.day);
+    }
+  }
+
+  if (dateParts && dateParts.year && dateParts.month && dateParts.day) {
+    return dateParts.year.concat('-', dateParts.month, '-', dateParts.day);
+  }
+
+  return '';
+};
```
petorsong: can this be moved into https://github.com/uwblueprint/paramedics-react/blob/master/src/utils/format.ts

src/components/Maps/Sidebar.tsx
```diff
@@ -134,31 +243,46 @@ const Sidebar = ({
...
           </Typography>
           <TextValidator
-            placeholder="Pin name"
+            placeholder={`${placeholderType} name`}
             onChange={handleLabelChange}
             value={label}
             validators={['required']}
-            errorMessages={['A pin name is required']}
+            errorMessages={[
+              `A ${placeholderType.toLowerCase()} name is required`,
```
petorsong: to avoid `A event name ...`
```suggestion
              `${placeholderType} name is required`,
```

src/App.tsx
```diff
@@ -28,23 +26,23 @@ function App() {
       <Route
         exact
         path="/events/new"
-        component={(props) => <EventCreationPage mode="new" {...props} />}
+        component={(props) => <MapPage mode="newEvent" {...props} />}
```
petorsong: these should use the enum strings as well (after importing `graphql/queries/maps`)
```suggestion
        component={(props) => <MapPage mode={MapModes.newEvent} {...props} />}
```

src/components/Maps/MapPage.tsx
```diff
@@ -288,51 +533,192 @@ const MapPage = ({
...
-        onSuggestionTempMarkerSet={onSuggestionTempMarkerSet}
-        setTempMarkerClick={() => {
-          setTempMarkerClick(false);
+      {sidebarRender && (
+        <Sidebar
+          open={openSidebar}
+          onClose={onSidebarClose}
+          mode={mode}
+          title={sidebarTitle}
+          clickedAddress={tempMarkerClick ? tempMarkerAddress : undefined}
```
petorsong, line 679: any values that are manually unset should be set to `null` rather than `undefined`
```suggestion
          clickedAddress={tempMarkerClick ? tempMarkerAddress : null}
```

src/components/Maps/Sidebar.tsx
```diff
@@ -134,31 +243,46 @@ const Sidebar = ({
...
+                      id="date-picker-inline"
+                      value={date}
+                      onChange={(date) => setDate(date)}
+                      KeyboardButtonProps={{
+                        'aria-label': 'change date',
+                      }}
+                    />
+                  </ThemeProvider>
+                </MuiPickersUtilsProvider>
+              </Container>
```
petorsong, line 258: this should be styled to match the name/location input fields that are also in the sidebar (size, font weight, etc.)

petorsong (dismissed):
looks good! just waiting for a couple more changes to display the static addresses for events/ccps on the auth/event dashboard

if you can tackle the comments i left that would be awesome, otherwise i can push it off onto another ticket

src/components/Maps/MapPage.tsx
aaronabraham311: This results in a Typescript error. If a prop is not mandatory, then it has two types: one which is the type of the object when passed (string, number, custom type) and the other type is `undefined`. Using `null` doesn't conform to the typing, hence why I use `undefined`.

petorsong (changes requested):
image

there's a case where `pinOfInterest` comes back as an empty array (which is not falsey) and causing this error - it's probably easily fixed by adding a check to see if the array is not empty as well as not null

aaronabraham311:
> there's a case where pinOfInterest comes back as an empty array (which is not falsey)

Just curious: do you remember how this ended up happening? Usually this is indicative of bigger problems so just wanted to make sure.

petorsong:
> Just curious: do you remember how this ended up happening? Usually this is indicative of bigger problems so just wanted to make sure.

i didn't check anything too deep, just logged `pinOfInterest` within that null check to see that it was sometimes and empty array (which threw the error) - it might be the case that the events corresponding to this issue are "legacy" ones that don't have a pin associated with them, which we should still be supporting since none of our existing events have associated locations yet

